PR #2, opened 2019-10-14
chore(.gitignore): Ignore *.pyc files

stephengtuggy:
For some reason, a bunch of .pyc files (compiled Python modules) got created in my Assets-Production folder at one point. This change .gitignore's them, so that they don't show up as files to check in.


Loki1950:
Yeah ignore *.pyc files is a good idea they will recreated with each run of vegastrike anyway

PR #8, opened 2020-07-06
Removes the debug warnings from the console

evertvorster:
This simple little edit hides the spammy debug error messages in the console

### DIFF
--- a/modules/ship_upgrades.py
+++ b/modules/ship_upgrades.py
@@ -168,10 +168,10 @@ def getItem (cat,parentcat=None):
     list=VS.getRandCargo(1,cat)#try to get a cargo from said category
     if (list.GetQuantity()<=0):#if no such cargo exists in this cateogry
         if (parentcat!=None):
-            debug.warn("Python Upgrade Error: finding %s using %s instead", cat, parentcat)
+#            debug.warn("Python Upgrade Error: finding %s using %s instead", cat, parentcat)
             list=VS.getRandCargo(1,parentcat)#get it from the parent category
         if (list.GetQuantity()<=0):#otherwise get cargo from upgrades category
-            debug.warn("Python Upgrade Error: category %s -- getting random instead", cat)
+#            debug.warn("Python Upgrade Error: category %s -- getting random instead", cat)
             list=VS.getRandCargo(1,"upgrades")#this always succeeds
     return list
 


evertvorster:
The messages are debug messages. There really should be a --debug switch that can be thrown to turn messages like this on. 

Still, for a production we don't need debug messages scrolling around the console obscuring real errors.

nabaco:
@evertvorster - is possible for you to add the switch, instead of just commenting the logs?
You could add just the logic for the switch, and we'll deal later with integrating it with our build system and/or engine.

evertvorster:
@nabaco, I am not a coder, and would not know where to start in making an ifdef for this. 

Since I have just commented out the debug message, it's easy enough for a troubleshooter in the future to re-enable to hunt down bugs.

My apologies for not being able to provide a nicer solution to this at this point in time.

BenjamenMeyer:
If they're just debug messages then we should send them to the log system at the debug level and the config will take care of it.

nabaco:
@BenjamenMeyer do you know how to do that in the python files?

BenjamenMeyer:
@nabaco not sure how it connects back to C/C++ land, but in Python:
```python
import logging

LOG = logging.getLogger(__name__)

LOG.debug("message")
```

If we could figure out how that connects with Boost and C++ to dump into the logs there that would be awesome. I'm sure there's a way, just haven't done it myself. Boost Python docs might inform us.

nabaco:
@stephengtuggy @BenjamenMeyer - so what's our verdict on this PR? @evertvorster clearly stated that he's not up for the job to make the switch?
Does one of you guys wish to pick this up?
If not, I say let's reject this PR, and open a ticket based on it, to be implemented in better times.

evertvorster:
@stephengtuggy OK, which is better... spamming the console with useless error messages making the game look unstable, or a nice calm console? Sure, a debug switch can be added, and we can stick that in as an issue on this Repo to be corrected, and reference this PR as a place to start. Would that not achieve both our goals?

stephengtuggy:
@evertvorster Hmm, I see your point.

At a minimum, I have filed #8 https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/167 and https://github.com/vegastrike/Assets-Masters/issues/13 . I guess that's good enough for now. I'm okay with merging this PR.

@BenjamenMeyer @nabaco @Loki1950 what do you think?

stephengtuggy:
@evertvorster this PR fixed https://github.com/vegastrike/Assets-Masters/issues/10 , right?

evertvorster:
Indeed it did. I closed that issue. :)

PR #9, opened 2020-07-11
Fixes the load errors. 

evertvorster:
This PR fixes the underlying cause of the Load Errors. 

Unfortunately, ~/.vegastrike/sectors need to be re-generated to get rid of the loading errors, as it is corrupted.

Simply remove/rename ~/.vegastrike/sectors, and Vega Strike will do the rest on the next run. 
This of course means that some installations move around a little in the Universe.

This has been tested against the unified units.csv PR that I am still waiting on, & works fine on that. 
I am able to load old save games without issue.

I am doing further testing, of course. 

I have done some limited testing against a split .csv, just for completeness. I won't recommend it. 
There is some weirdness, especially with older save games. 

There are still some un-related load errors that remain with the split .csv, as there are other files with miss-matching case variables that I can't edit. 
You might have to go around with sed -i and replace all the lower case base names with the correct case as they are laid out in this file in your .vegastrike/sectors if you want to bring over old save games, and fix the load errors in those. 

As far as I can tell the mechanism that introduces the load error is this:
When you start vegastrike with no ~/.vegastrike/sectors, vegastrike uses smallunits.txt to manually place installations in various universes. 
All these installations have their name in all small caps, whereas they are defined in 'Like_This' in units.csv, and in a similar way in the split .csv files.
When VS then wants to load these units, the case does not match up, and you get a LOAD ERROR. 
This smallunits.txt file seems to be used only when you go to a new place in the universe.

TL;DR: Every save file up to this fix is corrupted, and will have Load Errors until someone makes a tool to fix them.


nabaco:
Will merge after the unified CSV PR

PR #10, opened 2020-07-11
Removes the mis-aligned buttons in the nav console.

evertvorster:
Buttons are so last century.

It's only commented out in the file, so that we can make something prettier when we have the time.

The Nav computer needs an overhaul anyways. If I had a space ship that could jump to other galaxies,
and the nav computer installed in it was this basic I would give back that ship.

-------------------------------------------------------
I have a few ideas for improvements that should be fairly low-hanging. 

1. A search engine so you can map a route to any star that you have been to.
1.1 Route map to become an auto-generated mission?
2. "Fading" jump lines between solar systems depending on how far they are away from you to reduce on screen clutter
3. Un-fading mission routes to help with planning.

A screenshot of what the nav computer looks like without its buttons:
The horizontal line through is a glitch in my screen grabber util, it's not there in the game. 

Screenshot_20200711_103406

### DIFF
--- a/cockpits/nav/navdata.xml
+++ b/cockpits/nav/navdata.xml
@@ -4,6 +4,7 @@
 <Meshfile file="console.bfxm"    x_small=".12"  x_large=".7"  y_small=".15" y_large=".85" x_mesh_coord="-12.0" y_mesh_coord="0.0" z_mesh_coord="100.0" scale="1.4"/>
 </console>
 
+<!--
 <button1>
 <Meshfile file="b_generic.bfxm"   x=".755" y=".8"  dx=".11" dy=".05" x_mesh_coord="40.0" y_mesh_coord="42.0" z_mesh_coord="100.0" z_mesh_coord_delta="1.0" scale="1.4"/>
 </button1>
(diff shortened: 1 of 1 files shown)


nabaco:
Sounds good to me. Though this is more of a plaster to hide the issue, I think it looks good enough, and like @evertvorster said, we'll want to re-haul the UI either way.
I'm approving this.

nabaco:
@BenjamenMeyer  @stephengtuggy @Loki1950 what's your say on this?

BenjamenMeyer:
What functionality is being disabled?

Im okay with this on master, but we should have a replacement before releasing 0.7.x.

evertvorster:
@BenjamenMeyer, the whole thing functions. Originally, each button had a text label. I expect on a 1024x800 display the buttons might still be below the text. However, on other screen resolutions, the image of the buttons are against the side of the display, and not active. The text, in a fortunate coincidence, is still active. 

This PR removes the image of buttons, but leaves the click-able text behind.
The feel of the interface has changed from the manual pressing of buttons to clicking on hyper-links on a see-through touch sensitive screen.

PR #12, opened 2020-07-13
Units fixup

evertvorster:
This is an attempt to get older savegames to also not have loading errors. 

Just to be _very_ clear, this PR is not needed for newly created universes & savegames. 

I took every Base_Name in the units.csv file, made a duplicate with all smalls, eg: base_name. 
As you can see from the diff, the modding tools inserted these in the middle, making for quite a big diff.


nabaco:
That's an interesting one. I'll have to check it out locally to get a glimpse at the changes - Github refuses to show me.
@evertvorster could you please detail out what testing you did with this, with regard to savefiles?

evertvorster:
@nabaco, this is the process I used to hunt down where the LOAD Failed errors were coming from:

I would fly around until I saw a Load error, then note down the unit file name in the console. I could fix the load error by simply adding in the unit file name into the units.csv. Usually, there is a corresponding unit file name where the case of the characters in the name was different, which I would clone to make the new unit. 

Then I would grep through the contents of /usr/share/vegastrike to see where this variable was used/defined. It turns out that most of the missing ones were defined in universe/smallunits.txt (Just spotted another file with a wrongly-named Unit, smallunits.pirates.txt - expect a PR soon.)

When I updated the smallunits.txt to have the correct case for the unit names, the load errors disappeared on newly created universes. Unfortunately, it looks like the unit names are drawn into the universe you first generate when loading the game. 

So, there is the correct fix for this error, which is to make sure that the case of the various descriptions in "universe" is correct, and then there is the fix to not break the savegames, which you are looking at now. Let me show you a screen shot of the two files. My changes are on the right, highlighted.
image

PR #13, opened 2020-07-13
Resolution settings

evertvorster:
Adds 4K and near 4K resolutions. 
The near 4K resolutions are for windowed games.

Also moves the font_point to the resolution definitions. 
This has the effect that as you change resolutions, the fonts remain readable. 

The only drawback of point sizes larger than 16 causes fonts to overlap on the credits, new game and load game screens of the initial main_menu. I believe that this is a fair trade-off to having readable fonts on your HUD for the rest of the game.


nabaco:
The changes of the font settings make perfect sense :)
@evertvorster could you please share some screenshots of how the text looks now? Especially in the problematic places that you have mentioned?

evertvorster:
Sure thing, let me paste the screen shots here. 
I have tried to increase the spacing between text. The screens are defined in bases/main_menu.py. It looks like they are drawn on to a 1024x768 canvas, then scaled up to whatever screen size you have. 

Introduction:
image

Load Game Screen:
image

Funny thing is that the Credits screen has the proper spacing:
image

nabaco:
I see.
The credits and the "new mission" screen don't bother me as much, but the loading screen is an issue, as it seems to be less functional. Any chance you'll be able to fix that up?

stephengtuggy:
@nabaco your last comment on this PR has me a little worried. Is the loading screen still less than fully functional?

nabaco:
It's usable, just on higher resolutions the text overlaps a bit. If you that it affects functionality on your system substantially, we can revert this. Though I'm more inclined to keep it as it is, and try to find a fix for the issue.

PR #14, opened 2020-07-13
Smallunits.pirates & nebulae.txt

evertvorster:
I found some more files in universe that has smallcaps unit names. The names have been updated, but I do not know how to test these.
Is it possible to make a mission that contains nebulae and special units?

I'll experiment and come back. In the meantime, I'll leave this PR here.


nabaco:
As per vegastrike/Vega-Strike-Engine-Source#143 we're considering currently to remove Nebulas all together

PR #15, opened 2020-07-13
Updated .desktop file to run our vs script, wherever it may be.

evertvorster:
The current .desktop file just executes "vegastrike"
This tiny update lets it run "vs" which is a small bash script that calls vegastrike with the correct install location. 
Just a bit of polish, really.


nabaco:
Where is the script located?

evertvorster:
Currently it's in the root of Assets-Production for distribution.

The CMakeLists.txt will make an install target for make that will install it into your system's preferred binary directory. If you specify any switches on the command line for cmake, it will of course respect that. On my system the install directory is /bin. (/usr/bin just symlinks back to /bin on Arch) 

For you it might be /usr/local/bin. 

No matter where it goes, it just needs to sit somewhere in your PATH

PR #16, opened 2020-07-14
Updates the vs script to accept 1 argument, to load missions

evertvorster:
This enables the vs script to take 1 argument, to load missions.
Also, it now logs the console to a file & keeps 1 version of the old log

### DIFF
--- a/vs
+++ b/vs
@@ -1,15 +1,19 @@
 #!/bin/bash
+# This is a small wrapper that lets VegaStrike - Upon the Coldest Sea
+# run from seemingly anywhere
 set -e
 
 VEGASTRIKE_LOCAL_SHARE_DIR="/usr/local/share/vegastrike"
 VEGASTRIKE_SHARE_DIR="/usr/share/vegastrike"
 
-# This is a small wrapper that lets VegaStrike - Upon the Coldest Sea
-# run from seemingly anywhere
+#Let's keep a log of the console
+touch ~/.vegastrike/console.log
+mv -vf ~/.vegastrike/console.log ~/.vegastrike/console.bak 
+
 if [ -d "${VEGASTRIKE_LOCAL_SHARE_DIR}" ]; then 
-    vegastrike -d"${VEGASTRIKE_LOCAL_SHARE_DIR}"
+    vegastrike -d"${VEGASTRIKE_LOCAL_SHARE_DIR}" "$1" | tee ~/.vegastrike/console.log
 elif [ -d "${VEGASTRIKE_SHARE_DIR}" ]; then
-    vegastrike -d"${VEGASTRIKE_SHARE_DIR}"
+    vegastrike -d"${VEGASTRIKE_SHARE_DIR}" "$1" | tee ~/.vegastrike/console.log
 else
     echo "Unknown Game Asset Data Location."
     echo "Do you have the game assets installed?"
--- a/vs
+++ b/vs
@@ -8,12 +8,16 @@ VEGASTRIKE_SHARE_DIR="/usr/share/vegastrike"
 
 #Let's keep a log of the console
 touch ~/.vegastrike/console.log
-mv -vf ~/.vegastrike/console.log ~/.vegastrike/console.bak 
+touch ~/.vegastrike/console_err.log
+mv -f ~/.vegastrike/console.log ~/.vegastrike/console.bak 
+mv -f ~/.vegastrike/console_err.log ~/.vegastrike/console_err.bak
+echo "Starting Vega Strike, console output will be saved in ./vegastrike/console.log, errors in ~/.vegastrike/console_err.log"
+echo ""
 
 if [ -d "${VEGASTRIKE_LOCAL_SHARE_DIR}" ]; then 
-    vegastrike -d"${VEGASTRIKE_LOCAL_SHARE_DIR}" "$1" | tee ~/.vegastrike/console.log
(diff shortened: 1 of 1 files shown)


vs
```diff
@@ -1,15 +1,19 @@
...
+# This is a small wrapper that lets VegaStrike - Upon the Coldest Sea
+# run from seemingly anywhere
 set -e
 
 VEGASTRIKE_LOCAL_SHARE_DIR="/usr/local/share/vegastrike"
 VEGASTRIKE_SHARE_DIR="/usr/share/vegastrike"
 
-# This is a small wrapper that lets VegaStrike - Upon the Coldest Sea
-# run from seemingly anywhere
+#Let's keep a log of the console
```
nabaco, line 9: off?

vs
```diff
@@ -1,15 +1,19 @@
...
+touch ~/.vegastrike/console.log
+mv -vf ~/.vegastrike/console.log ~/.vegastrike/console.bak 
+
 if [ -d "${VEGASTRIKE_LOCAL_SHARE_DIR}" ]; then 
-    vegastrike -d"${VEGASTRIKE_LOCAL_SHARE_DIR}"
+    vegastrike -d"${VEGASTRIKE_LOCAL_SHARE_DIR}" "$1" | tee ~/.vegastrike/console.log
```
nabaco, line 14: wouldn't `tee` cause it to be on the screen too?
either way, can we do this both for `stdout` and `stderr`?

vs
```diff
@@ -1,15 +1,19 @@
...
+# run from seemingly anywhere
 set -e
 
 VEGASTRIKE_LOCAL_SHARE_DIR="/usr/local/share/vegastrike"
 VEGASTRIKE_SHARE_DIR="/usr/share/vegastrike"
 
-# This is a small wrapper that lets VegaStrike - Upon the Coldest Sea
-# run from seemingly anywhere
+#Let's keep a log of the console
+touch ~/.vegastrike/console.log
```
nabaco, line 10: Why create a file and then move it? shouldn't it be other other way around?

nabaco, line 10: Why create a file and then move it? shouldn't it be other other way around?

vs
evertvorster: Yes, it leaves the console output to the screen, if there is a screen attached. We can modify it for stdout and err, if you want. Right now it captures _everything_ in it's glorious completeness.

vs
evertvorster: The idea here is to move the previously captured log out of the way, so it does not get clobbered instantly. 
mv cannot fail graciously if the file it wants to move does not exist. 
There are a whole host of reasons for this file to not exist. 

If the source file for mv does not exist, it fails and the script exits. 
The main purpose of touch is to update the "modified" time of an existing file. Creating the file if it does not exist is a happy side-effect. So, now if the file does not exist, "touch" makes it, otherwise it harmlessly updates it's modified time, allowing the mv command to not fail, as it is guaranteed to have a file to move.

I could have implemented similar logic with:
if -e ~/vegastrike/console.log; then ...
    mv foo bar
fi

However, I nearly always get my if tests wrong in bash, and then it's a bit of back and forth until it works right.

vs
evertvorster: The tee command writes it's input to the console & to the file specified. 
I don't want to fill the .vegastrike directory with files for people that are not interested, so my log rotation is using 2 files only.

vs
nabaco: let's have it with an if command
```
if [ -f ~/vegastrike/console.log ];
then
    mv foo bar
fi

File: vs [hunk 4]
```diff
@@ -1,15 +1,23 @@
...
+touch ~/.vegastrike/console_err.log
+mv -f ~/.vegastrike/console.log ~/.vegastrike/console.bak 
+mv -f ~/.vegastrike/console_err.log ~/.vegastrike/console_err.bak
+echo "Starting Vega Strike, console output will be saved in ./vegastrike/console.log, errors in ~/.vegastrike/console_err.log"
+echo ""
+
 if [ -d "${VEGASTRIKE_LOCAL_SHARE_DIR}" ]; then 
-    vegastrike -d"${VEGASTRIKE_LOCAL_SHARE_DIR}"
+    vegastrike -d"${VEGASTRIKE_LOCAL_SHARE_DIR}" "$1" --debug=3 >&~/.vegastrike/console.log 2>~/.vegastrike/console_err.log
```
nabaco, line 18: are we sure we need the `&` symbol here?

nabaco, line 18: are we sure we need the `&` symbol here?

stephengtuggy:
When https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/172 is ready, it should take care of the logging stuff. Most of it, anyway.

stephengtuggy:
Looks good though, as far as running Vega Strike with the right data directory, etc.

evertvorster:
OK, I implemented the if statements. 
When the vs script is run, it tells you where it is saving the logs.

Of course, ~./vegastrike is NOT the proper place for logs, that would be /var/log. I'm taking a little liberty here as Vega Strike is not exactly system software... Which is why I am also overwriting the log on every run, and keeping just 1 backup. 

All this is very Linux specific. I read about Windows having some sort of Linux support these days. Has anyone tested it?

nabaco:
> All this is very Linux specific. I read about Windows having some sort of Linux support these days. Has anyone tested it?

Not sure what your mean. If you're talking about WSL, it's irrelevant in our case I think.

Either way, windows compilation is still WIP, and its installation and startup works differently, so we're fine with this changes

PR #18, opened 2020-07-15
Licences

evertvorster:
This HTML describes the licenses, and provides handy links for as long as the Wiki stays up.
Since we want to keep the description of the acceptable license for art with the code, distributing it the the package is the next logical step.


nabaco:
I think we should have a more developer friendly version too. You can run the document through Pandoc to receive a markdown version of this.
Can be done in a separate PR if you wish

BenjamenMeyer (changes requested):
Mark Down (MD) or ReStructured Text (RST) would be greatly preferred.

royfalk:
I like this, but I think we should have a review of the actual content.
It's hard to review like this.
@evertvorster , can you extract this to a discussion?

BenjamenMeyer:
> I like this, but I think we should have a review of the actual content. It's hard to review like this. @evertvorster , can you extract this to a discussion?

Markdown (MD) or ReStructured Text (RST) would resolve that since GitHub can render those.

royfalk:
Is this still relevant? 
@evertvorster can you ask the AI to move this to an md file?

BenjamenMeyer:
> Is this still relevant? @evertvorster can you ask the AI to move this to an md file?

I do think we need to update the `LICENSE` file and other in-repo documentation to discuss our more nuanced approach for licenses in this repository. Whether this is the right PR for that I'm not sure... I couldn't really read the HTML stuff since Github doesn't render that nicely.

evertvorster:
File is markdown now, should render properly in Github

documentation/Development_Graphics_Requirements.md
```diff
@@ -0,0 +1,296 @@
...
+
+- DQ - Development Quality: textures with very low horizontal resolution and low degree of artistic quality
+
+- RQ - Release Quality: textures with at least medium horizontal resolution and medium to high degree of artistic quality
+
+- CQ - Cinematographic Quality: textures with high horizontal resolution and very high degree of artistic quality
+
+Specific resolution requirements can be found on the development pages specific to each image type.
+
+## SVN Structure
```
BenjamenMeyer, line 222: We probably need to update this to reflect Git instead of Subversion.

documentation/Development_Graphics_Requirements.md
```diff
@@ -0,0 +1,296 @@
...
+
+- (CC-SA) Creative Commons Share Alike license
+
+- (CC-BY-SA) Creative Commons By Attribution Share Alike license
+
+Please note that we do not allow licenses not mentioned above, in particular:
+
+- (CC-NC) Creative Commons Non-Commercial license or any combination with CC-BY and CC-SA
+
+What this means is that, besides the licensing of your work, you need to submit/share the source files, and in terms of artwork, the master/project files for your contributions. If you are not willing to contribute under one of the accepted licenses then simply forget it.
```
royfalk, line 27: How about:
What this means is that, besides the licensing of your work, you **must** to submit/share the source files, and in terms of artwork, the master/project files for your contributions.

documentation/Development_Graphics_Requirements.md
```diff
@@ -0,0 +1,296 @@
...
+- Cargo images
+
+- Space backgrounds
+
+- Animation images
+
+# Graphic Files Requirements
+
+## Overview of Graphics Requirements
+
```
royfalk, line 60: I'm not sure this is still relevant but it surely won't be for long.

documentation/Development_Graphics_Requirements.md
```diff
@@ -0,0 +1,296 @@
...
+- 64 px
+
+- 128 px
+
+- 256 px
+
+- 512 px
+
+- 1024 px
+
```
royfalk, line 102: Can't we go higher today?

documentation/Development_Graphics_Requirements.md
```diff
@@ -0,0 +1,296 @@
...
+- Animation images
+
+- Planet textures
+
+In case of doubt please ask one of the developers or on the forum.
+
+### Compression with nvcompress
+
+Please note that nvcompress (all versions) is invisibly corrupting DXT5 compressed textures for older nVidia graphic cards principally with drivers before version 169.09. Please update to the latest drivers for testing dds textures.
```
royfalk, line 161: Is this still true today?
Version 169 was released in 2007, 20 years ago.

documentation/Development_Graphics_Requirements.md
```diff
@@ -0,0 +1,296 @@
...
+
+You will need nVidia's free texture tool nvcompress to transform your original textures to optimized dds textures.
+Get the tool here: [NVIDIA Texture Tools](http://developer.nvidia.com/object/texture_tools.html)
```
royfalk, line 164: This leads to legacy tools. The updated tool is https://developer.nvidia.com/texture-tools-exporter

documentation/Development_Graphics_Requirements.md
```diff
@@ -0,0 +1,296 @@
...
+
+**Sidenote**: The following applies only to NVIDIA texture tools version 0.9.4. More recent versions exist. Due to a bug in handling 1 pixel mipmaps in the original (0.9.4) version, you will be further required to patch the tools with safemode's patch. The patch can be obtained here: [save nvidia-texture.patch file](http://signal-lost.homeip.net/files/nvidia-texture.patch). Patch the texture tools, compile, and install them.
```
royfalk, line 166: Probably also not relevant.

documentation/Development_Graphics_Requirements.md
```diff
@@ -0,0 +1,296 @@
...
+
+- `*nvcompress -bc2 (-nomips) texture_original.png texture_dds.image*`
+
+For DXT5 images with smooth transparency gradient:
+
+- `*nvcompress -bc3 (-nomips) texture_original.png texture_dds.image*`
+
+### Compression with Gimp dds plugin
+
+Gimp plugin produces dds images with lower quality than that one produced by the nvcompress tool. In addition, the plugin uses hardware compression and may produce different results on different systems. Therefore, compressing images with this plugin is not recommended for submission, but can be used as an alternative method for local testing purposes only.
```
royfalk, line 188: May not be relevant today.

documentation/Development_Graphics_Requirements.md
```diff
@@ -0,0 +1,296 @@
...
+
+- copyright/copyleft information should go into: *imagefile_license.txt*
+
+- source/project file naming: *Source images should be*imagefile_source.xcf*(or whatever extension)*
+
+- if several source files are required to execute the project, place them in a zip file
+
+# The Selection and Vetting Process
+
+When submitting new art, it is recommended to request feedback from the community through the forum before submitting the images or textures.
```
royfalk, line 252: Not sure there is a community to get feedback from today.

royfalk (commented):
It feels like most of this document is now stale. Not sure if you want to deal with this, but if you do, I left some helpful comments.

documentation/Development_Graphics_Requirements.md
BenjamenMeyer: yeah, this can probably use an update.  @evertvorster where did the source for this come text come from? wiki.vegastrike.org?

documentation/Development_Graphics_Requirements.md
BenjamenMeyer: We should be able to; though we should probably specify a ceiling but mostly for storage purposes to keep the repo and releases from getting too large.

Though we can probably solve the release issue if we split the package up into:
- OS Specific Release Packages
- Data Resource Packages

where the OS Specific Release Packages do the heavy lifting and get all the stuff required to make the game functional on specific operating systems; and then the user could be provided the option to download the Data Resource Packages for the resolution level they choose. So if they want hi-fidelity audio they would have to download a hi-def audio package; if they want 4K images they could download the 4K image package. We would have to provide some kind of downloader/extraction tool in the settings though...  So probably something for the future... but it could help save space since we'd only have 1 copy instead of several.

evertvorster:
This was a very old request in the first place. 

Give me the overall idea of how you would like this document to look, or at least the constraints you have in mind. 

I found with my work on the fork that there is a whole host of different licenses on various parts of the code, and I'm sure that there is a similar situation on the art. 

Should I mine what we distribute for licenses already in use, and then put a suggested license for people who want to contribute art but is not sure which one to pick? 

We'll focus this document on licenses only, and leave the modding stuff for how-tos and the wiki, and just link to the licence doc when the subject comes up elsewhere?

documentation/Development_Graphics_Requirements.md
evertvorster: True.. I just converted the doc without reading it.

documentation/Development_Graphics_Requirements.md
evertvorster: Agreed

documentation/Development_Graphics_Requirements.md
evertvorster: Yeah, art has been stationary for a while. I'm thinking of making shaders available for the skyboxes and planet surfaces, and that is a form of art, so there will be at least one person fiddling with the art in the future.

documentation/Development_Graphics_Requirements.md
evertvorster: I highly doubt it. I'll play with the tools, but should this really be in a licenses document?

documentation/Development_Graphics_Requirements.md
evertvorster: It was the wiki article, if I remember correctly, this request is nearly six years old by itself... Pretty soon it will be old enough to vote!

documentation/Development_Graphics_Requirements.md
evertvorster: That is a good suggestion, however totally un-enforceable. 
I would go more for something like this: "If you make the sources available to re-create your art, when technology changes in the game, your art will have a much better chance of being included in future versions of Vega Strike."

documentation/Development_Graphics_Requirements.md
evertvorster: Yeah, I doubt anyone uses subversion anymore

documentation/Development_Graphics_Requirements.md
BenjamenMeyer: @evertvorster thanks for the context. we probably need to bring it up-to-date at a minimum; but we also don't want to require brand new PCs to run the game (my newest is 2022 era; but most of mine are way older; I don't upgrade computers often).

documentation/Development_Graphics_Requirements.md
BenjamenMeyer: > I highly doubt it. I'll play with the tools, but should this really be in a licenses document?

probably not. This kind of information is more for the README and HOWTO documentation. The LICENSES document should focus on the licenses that everything is distributed under and any rules we may have about licensing.

BenjamenMeyer:
@evertvorster let's open a discussion on this topic

PR #22, opened 2020-07-17
Load failure fix2

evertvorster:
This one character PR disables the loading of bases with broken names completely. I've done the two Janus missions, and am halfway to Sol. 

So far I prefer this more sparsely-populated universe. There is usually a base or two in every system, so it's not like there's nothing out there. 

During my testing of trying to fix the names, I once had a Stirling with three (3) identically named Commerce Centers, four(4) refineries and a shipyard and a couple of relays. 

That's just too much! - Too much of a good thing is not a good thing. 

I'll leave this PR here while I do further testing on it, case anybody else wants to test this Universe where everything has a name, and there are no load errors.
This fix does not affect old save games, as the bases only gets generated when you go to a new system.


modules/generate_dyn_universe.py
```diff
@@ -119,7 +119,7 @@ def AddBasesToSystem (faction,sys):
         tn =[]
         for i in range(len(shiplist)):
             tn+=[ (shiplist[i],nums[i])]
-        fg_util.AddShipsToFG(fg_util.BaseFGInSystemName (sys),faction,tn,sys)
+#        fg_util.AddShipsToFG(fg_util.BaseFGInSystemName (sys),faction,tn,sys)
```
stephengtuggy, line 122: So this code no longer generates any bases at all? It's left to some other code elsewhere to do it?

BenjamenMeyer: It might be improperly loading the systems and not properly detecting that they've already been loaded

evertvorster: There seems to be two parallel functions for creating bases. 
This one can't name the bases properly, and sometimes ends up making up to seven bases in a system. 
The bases that are generated with the other function end up in .vegastrike/sectors, work properly, and seem to be the right amount of bases, for me at least.

stephengtuggy: I think it would be a good idea to have some other people play-test as well, before merging.

evertvorster: @stephengtuggy I fully agree with you, this does need a lot of play testing. You simply can't just take my word for it that the gameplay is better or worse. What you enjoy in a video game could be completely different that what I would like. 
Let's consider this PR. All it does is comment out a command that is causing issues. To test it you could very easily do the same edit on your test machine and go for a spin in the VS universe.
You could just accept the PR and have the same effect, and if you hate it you could roll it back. Or, make a single character edit and be back to the old behavior. 
I have a serious issue with the old behavior, in that the names of the installations is corrupted. This is still something I want to fix. When I do crack the nut, I would like to make this a configure option for users to turn on or off as they like. 
In the meantime, I'm zooming around in a VS universe that has proper names everywhere, and the relative sparseness of land-able bases makes the user jump around more, and get to explore more. I've been playing in this type of universe for the last few hours, and I quite like the experience. 

I would like you all to test this to see if you agree, please.

stephengtuggy:
I believe this PR has been superseded by #25 .

evertvorster:
@stephengtuggy Yes, you are right, I'll close this sucker down

modules/generate_dyn_universe.py
BenjamenMeyer: Regarding play testing: there's a way to checkout PRs locally to build them (see https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally); alternatively one could pull down anyone else's fork either directly or via an additional remote.

There's some cases like this where it will be necessary to do those kinds of things. Hopefully as we get better understanding of the code and can mature stuff for maintenance we can make those the exception and not the rule.

PR #23, opened 2020-07-18
Flightgroup names2

evertvorster:
This one replaces #22 and #17 

I've removed the non-ASCII characters from ISO.txt, and removed it's duplicates in universe/fgnames.txt
The only reason I've left in ISO is so that you can compare it with universe/names.txt
There are some differences in the names, with universe/names.txt having a few more names, thus being more complete.

Spawner.py calls fg_util.py with the faction name, and if not found in universe/fgnames will use universe/names.txt
This file has all the human-sounding names, and a few more than the factions did. 

What is becoming clear to me is that in the Vega Strike universe, the humans kept their naming convention, and the other races have other ways of naming things. 
This means we'll probably have to duplicate rlaan to rlaan_briin to have a similar effect for that faction of the rlaan race.


evertvorster:
I believe I mentioned in the PR that all the human factions are using exactly the same name file. 

Like exactly.

All these duplicates are derived from universe/names.txt, as in universe/names.txt has all the names of the factions, and a handful more. 

Why keep around 10 copies of the same file? Especially if we are going to distribute it. 

I would understand if the names inside each file was different, for instance if the pirates had pirate-sounding names. I can also imagine a place where we could have a base name file, and then concatenate it with the extra names from a faction, then randomly pick from that bigger list. This would be for some future development, I suppose. 

In my play testing so far, all the human factions have human names, and all the other races have their own unique sounding names on installations and craft.

BenjamenMeyer:
@evertvorster thanks for the explanation. I had noticed the duplication too and quite agree. Just wanted to make sure we we're dropping anything that was needed. Thanks!

PR #24, opened 2020-07-20
Extra base names

evertvorster:
Leeedle bit of hacking to get the extra bases named. 

Unfortunately the function was designed to make either planets, nebulae or flight groups, despite it's "Jumppoint" name. 
So, the extra bases are added in as flight groups of 1 each. They are not added into sectors, and are transient.

In the game universe their transient-ness might be due to them being in a flight group, as they are not permanent installations like the bases the game makes. It's not inconceivable that a faction might want to move bases to new territory. 

Each of the bases created by this function has =0= in their name. They are always created after all the other flight groups, so are easily accessible with "n" then "shift t" a few times.

They are dock-able and are named correctly, which solves my problem of having units floating around space that are not named. 
To get them to appear, un-do PR #22 which still needs to turn into a configuration option, but for now there is no harm in having a few extra transient bases floating around the universe as a default. 
I'm play-testing this one now, having visited a handful of systems I have not noticed anything untoward.


stephengtuggy:
I think these bases need to be accessible with "n" directly. "n" should cycle through all significant objects, including transient bases. IMHO, at least.

evertvorster:
I am backing out this PR in favor for a more elegant fix provided by @s0600204

PR #36, opened 2020-10-11
chore: Omit Fedora 30 from Travis builds (master)

stephengtuggy:
Since Fedora 30 is End of Life now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ dist: focal
 # to the matrix to cover the distro
 jobs:
     include:
+        # Just build one .deb package for all .deb-based platforms
         # - os: linux
         #   name: "Ubuntu 20.04 focal"
         #   env:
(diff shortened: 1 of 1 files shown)


Loki1950 (approved):

.travis.yml
```diff
@@ -65,10 +66,6 @@ jobs:
           name: "Fedora 31"
           env:
             - FROM='fedora:31'
-        - os: linux
-          name: "Fedora 30"
-          env:
-            - FROM='fedora:30'
```
BenjamenMeyer, line 71: We should add newer versions when removing older versions.

stephengtuggy: Is Fedora 33 out yet? It doesn't look like it: https://fedorapeople.org/groups/schedule/f-33/f-33-key-tasks.html

Loki1950: Fedora likes to only keep 2 active versions the with the upstream repo always called Rawhide

BenjamenMeyer: okay - so we should only have 2 active versions in support at any given time. I don't use Fedora myself so just trying to make sure we're keeping things in the supported versions appropriately. Going forward, I'd expect we drop one and add one at roughly the same time.

Loki1950: Fedora is Red Hat's experimental test bed what works gets added to official RH releases and then the Centos project re-brands that

PR #38, opened 2020-10-31
bases/main_menu.py: Bump version in intro title and credits title to 0.7

stephengtuggy:
The 0.6.2 version of vsUTCS still says 0.5.2 alpha in these two places. Oops!

Updating master to reflect version 0.7.0. We may want to update the 0.6.x branch as well.

### DIFF
--- a/bases/main_menu.py
+++ b/bases/main_menu.py
@@ -48,7 +48,7 @@ def enterCredits(self,params):
     global plist_credits
     VS.musicPlayList(plist_credits)
 
-credits_title = """\t=== Vega Strike 0.5.2 alpha ===
+credits_title = """\t=== Vega Strike 0.7.0 ===
 \t        ---Credits---"""
 
 credits_text_col1 = """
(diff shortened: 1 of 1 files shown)


Loki1950 (approved):
Nice catch @stephengtuggy

royfalk:
At work, the way we avoid issues like this is by having the CICD plug in the value from the branch name. Something to consider.

BenjamenMeyer:
@royfalk its mixed. There's a few solutions methods; I haven't found a good one yet. Using branch/tag names makes it hard to use `0.6.x` as the branch as `x` isn't valid in the numeric field, meaning we're also having to make a lot more branches.

Other solutions use environment variables, or specialized packages. Honestly I haven't seen a good one. We can certainly explore options in an Issue though.

Loki1950:
What about adding a check and verify step to the release routine for now so we at least don't get to embarrassed in the future.

BenjamenMeyer:
For my Python projects I usually add something to the unit tests to make sure things match up (typically b/c it's specified in multiple places)... not sure how we can do that here or if that would apply.

Opened https://github.com/vegastrike/Assets-Masters/issues/19 to continue the discussion.

PR #42, opened 2020-12-28
Icon

evertvorster:
Added an icon for the setup. 
Added a .desktop file for vegasettings
Added both of those to the CMakelists.txt


stephengtuggy (approved):
Looks good to me! Nice icon.

PR #45, opened 2021-01-14
Migration: Drop Support for Travis-CI

BenjamenMeyer:
Thank you for submitting a pull request and becoming a contributor to Vega Strike: Upon the Coldest Sea.

Please answer the following:

Code Changes:
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation
- [ ] This is a documentation change only
- [x] CI System Change

Issues:
- #43, #44

Purpose:
- What is this pull request trying to do? 0.8.x
- What release is this for? 0.8.x
- Is there a project or milestone we should apply this to? 0.8.x

Issue #43, opened 2021-01-02
Move from Travis-CI to GitHub Actions


BenjamenMeyer:
This is a the Assets Production version of 
https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/382.

BenjamenMeyer:
Started porting over the functionality from the engine for master. Once that's ported to master we can then port it to 0.7.x.

stephengtuggy:
We never finished this on master, did we? 0.7.x has more recent changes than master does, on the GH-Actions-related files.

BenjamenMeyer:
@stephengtuggy not sure off hand, an audit is easy enough to do

stephengtuggy:
I did an audit. The files on master still reference `TRAVIS_TAG` and `TRAVIS_OS`. We'll need to fix that.

BenjamenMeyer:
@stephengtuggy thanks. I'll put together a PR tonight.

stephengtuggy:
@BenjamenMeyer Sounds good.

BenjamenMeyer:
@stephengtuggy sorry for the delay - https://github.com/vegastrike/Assets-Production/pull/54

Loki1950:
Life happens just reviewed #54

PR #46, opened 2021-01-14
Migration 0.7.x gh actions

BenjamenMeyer:

Please answer the following:

Code Changes:
- [x] CI System Change

Issues:
- #43
- #44 
- #45 

Purpose:
- What is this pull request trying to do? Port of PRs #44 and #45 to 0.7.x
- What release is this for? 0.7.x
- Is there a project or milestone we should apply this to? 0.7.x


.github/workflows/gh-actions-release.yml
```diff
@@ -0,0 +1,79 @@
...
+        FLAGS:      ${{ matrix.FLAGS }}
+        MY_OS_NAME: linux
+        IS_RELEASE: 1
+      run: script/cibuild $FLAGS
+    - name: Upload the artifacts
+      uses: skx/github-action-publish-binaries@master
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        args: 'packages/*'
```
stephengtuggy, line 79: We should probably leave out the source packages, in similar fashion to https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/402 , so that we don't end up with file conflicts during the release process.

stephengtuggy (commented):
Looks good overall. The only thing we might want to change is the build artifact pattern, in order to avoid file conflicts during releases.

.github/workflows/gh-actions-release.yml
BenjamenMeyer: I'll have to add that to Master as well.

.github/workflows/gh-actions-release.yml
```diff
@@ -0,0 +1,79 @@
...
+        FLAGS:      ${{ matrix.FLAGS }}
+        MY_OS_NAME: linux
+        IS_RELEASE: 1
+      run: script/cibuild $FLAGS
+    - name: Upload the artifacts
+      uses: skx/github-action-publish-binaries@master
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        args: 'packages/Vega-Strike_v*.*'
```
stephengtuggy, line 79: See if you can do something like this:

```yml
with:
  args: [ 'packages/*.rpm', 'packages/*.deb' ]
```

stephengtuggy: The base filename would be `vsUTCS_v*.*` in this case. And that won't exclude the source ZIP, the way it does for the engine, because of the different CPack naming strategy.

BenjamenMeyer: Unfortunately I don't think that'd work as according to https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepswithargs the `job.jobid.steps.with.args` doesn't support arrays of strings, just a single string. :disappointed:

BenjamenMeyer: I propose we table the sub-selecting for the moment. We can fix that in another PR.

stephengtuggy: OK

stephengtuggy: Oh wait. Why not just repeat the entire "Upload artifacts" step, only with a different wildcard pattern? I.e.,

```yml
    - name: Upload the artifacts - .deb
      uses: skx/github-action-publish-binaries@master
      env:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
      with:
        args: 'packages/vsUTCS_v*.deb'
    - name: Upload the artifacts - .rpm
      uses: skx/github-action-publish-binaries@master
      env:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
      with:
        args: 'packages/vsUTCS_v*.rpm'
```

stephengtuggy: If that doesn't work, or if you want to wait to try it in another PR, then that's fine, and I'll approve this PR.

BenjamenMeyer: I'd rather do that in another PR where we can explicitly experiment with it. That way the work on that is focused rather than appending multiple commits to this one to try things and then fix when GH complains or it doesn't work as expected.

BenjamenMeyer: It may also be best to try that on the Engine since the releases here are significantly larger.

stephengtuggy: OK, no problem. Approved then.

stephengtuggy (approved):
LGTM

PR #59, opened 2021-05-23
Convert units.csv to units.json and script

royfalk:
**This should not be merged.** 

I have not tested this, though this should not impact gameplay or even be merged at this point. 
This is a required step before changing the code to read from a json instead of a csv file. 
Note this also exposed at least one case where a ship had its upgrades in the wrong (lights) column.
Once we do this, we can also merge the weapons.xml into it and split it into files by faction and type.


Loki1950 (commented):
Sorry Roy but I do not feel competent to review this

royfalk:
@Loki1950 you can absolutely review the units.json file. The script is what's converting the units.csv to units.json. The actual content is the a json with the key as the header column and the value as the corresponding cell. I've also removed the empty cells, which makes it easier to spot problems than in the original csv file.
It's a huge file, so if you are reviewing, please add a comment of which units you covered (starting and ending).

Loki1950:
Have to clone it locally first will do that later this afternoon,your comment about the empty cells is valid as that is one of the major issues with a human understanding it

Loki1950:
Looking though units.json nowhere do I find reference to the actual texture file names the mesh files are there and the HUD sprites but not the ingame models textures

ministerofinformation:
Would it be possible to think about a tool to automatically generate a set of warnings about the current data set in a lint-like fashion (validator would be even better, but probably not plausible) for the resultant json file as we're considering the conversion?

royfalk:
> Would it be possible to think about a tool to automatically generate a set of warnings ... (validator would be even better, but probably not plausible) 
Ah. One of the nice things about JSON (and XML) is the schema. We can generate a schema for enforcing correctness of unit files. See https://json-schema.org/ and https://www.jsonschemavalidator.net/ .
Moving away from csv makes the format much more verbose (all those keys) but also makes it much more readable and easy to edit.
Also, it makes more sense to split the units file into types, each with its own schema: Ships, radars, etc. That would also mean compatibility of components to ships could be handled by tags (e.g. Aeran) and not just size. 
But I'm getting ahead of myself - we're 6-12 months away from this.

stephengtuggy:
> > Would it be possible to think about a tool to automatically generate a set of warnings ... (validator would be even better, but probably not plausible)
> 
> Ah. One of the nice things about JSON (and XML) is the schema. We can generate a schema for enforcing correctness of unit files. See https://json-schema.org/ and https://www.jsonschemavalidator.net/ .
> Moving away from csv makes the format much more verbose (all those keys) but also makes it much more readable and easy to edit.
> Also, it makes more sense to split the units file into types, each with its own schema: Ships, radars, etc. That would also mean compatibility of components to ships could be handled by tags (e.g. Aeran) and not just size.
> But I'm getting ahead of myself - we're 6-12 months away from this.

That sounds great, @royfalk . Maybe we could develop the schema, at least, sometime in the next month or two. I think that would be a good start.

BenjamenMeyer:
Not sure we could get an XML schema validator...IIRC some of the XML isn't quite to spec; but yes moving into that kind of things and fixing things up so we can use standardized tools to do it would be great.

units/parser.py
```diff
@@ -0,0 +1,133 @@
...
+            
+            parseLine(line)
+            #break
+        
+        line_num +=1
+       
+json_object = json.dumps(units, indent = 4)
+with open("units.json", "w") as json_file:
+    json_file.write(json_object)
+print(json_object)
```
BenjamenMeyer, line 131: This is a great first round; the globals could be avoided by using a Python Object structure instead of being purely functional design like it is now, but we can do more of that later.

BenjamenMeyer (commented):
Not sure I can do the CSV to JSON validation myself....the Python side looks good. I'd rather see a class/object structure used instead of the functional design but this is a good start and we can always address that later.

If someone can sign off on the JSON data-side then we're good to go:
- [x] Python approved
- [x] JSON approved

royfalk:
This is now ready to merge. It must be paired with https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/686.

units/how_to_convert_to_json.txt
```diff
@@ -0,0 +1,7 @@
+A brief guide on how to convert to JSON
+=======================================
+1. Open the CSV file in some kind of CSV editor (not a text editor).
+2. Search and replace all commas in the description with temporary, unused character. I used pipe. 
+3. Save the file.
+4. Run python3 parser.py
+5. (Optional) Open the units.json file in a text editor. Search and replace all pipes back to commas.
```
BenjamenMeyer, line 7: we could probably avoid this by using `"` around all the CSV values and escaping the `"` within the strings; then we wouldn't have to do steps 2 and 5.

BenjamenMeyer:
Took a look at a few of the JSON entries; and good enough for a first pass. I'm curious how some things are going to map out since the CSV data seems like it might be referencing other CSV files from within CSV entries, and we'll need to mitigate that somehow going forward, but it's a good first pass nonetheless.

I'd like to see things broken into sub-structures, for example:
```
{
    ...
    Armor_Front_Top_Right...
    Armor_Front_Top_Left...
    Armor_Front_Bottom_Right...
    ...
}
```
This should eventually move to something like:
```
{
     Armor: {
        Back: {
            Bottom: {
                 Left:..
                 Right:...
            },
            Top: {
                 Left:..
                 Right:...
            }
        },
        Front: {
            Bottom: {
                 Left:..
                 Right:...
            },
            Top: {
                 Left:..
                 Right:...
            }
        }
     }
```
But that doesn't need to be done in this PR.

BenjamenMeyer (approved):
Nothing blocking; looks good pending further play testing.
If we can switch between CSV and JSON easily, then we should merge so folks can test using the master branch, and then we can start rolling it out with the releases.

units/how_to_convert_to_json.txt
royfalk: To clarify, the issue is that the CSV file has commas in the textual description of the unit. As a result, the game itself parses the CSV by ignoring everything in the description. Strangely enough, LibreOffice Calc is smart enough to figure it out as well. 
Once we parse it and convert it to JSON, the JSON API in either C++ or python is smart enough to deal with anything (obviously).

BenjamenMeyer: @royfalk yes, one gets around the commands in the field by quoting the fields per the CSV specification (same with TSV and tabs in the field).

PR #60, opened 2021-05-26
Utf8 check

P-D-E:

Please answer the following:

Code Changes:

Issues:
- A similar call to loadGame is in main_menu.py, ~~still to be investigated~~ it's called when starting a new game (Campaign button), no further action needed.
- The in-game menu available when docked is handled by `basecomputer.cpp` in `cmd`, so the cpp check on the engine side is still needed
- The related github issue is the [#149](https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/149) in the source repository.

Purpose:
- What is this pull request trying to do?    **Preventing crashes from loading non-utf8 saved games via pre-emptive checks, alerting the user when it happens.**
- What release is this for?    **0.8.x**
- Is there a project or milestone we should apply this to?     **0.8.x**

### DIFF
--- a/bases/computer_lib.py
+++ b/bases/computer_lib.py
@@ -7,6 +7,7 @@ import VS
 import ShowProgress
 import methodtype
 import mission_lib
+from utf8_check import is_utf8
 
 pirate_bases = {
 }
@@ -216,7 +217,16 @@ class QuineComputer:
                 ShowProgress.activateProgressScreen('loading',3)
                 import dj_lib
                 dj_lib.enable()
-                VS.loadGame(self.picker_screen.items[self.picker_screen.selection].data)
+                game = self.picker_screen.items[self.picker_screen.selection].data
+                if not is_utf8(game):
+                    alert = f'  {game} is not UTF-8, convert it before loading'
+                    print(alert)
+                    self.alert_text = GUI.GUIStaticText(self.guiroom, 'txt_screen', alert,
+                        GUI.GUIRect(60, 90, 657, 100), color=GUI.GUIColor(255, 255, 255), bgcolor=GUI.GUIColor(0, 0, 0))
+                    self.alert_text.show()
+                    return
+                else:
+                    VS.loadGame(game)
         elif button_index == "btn_save":
             if self.mode != button_index:
                 self.picker_screen.items = [GUI.GUISimpleListPicker.listitem("New Game",NewSaveGame)]+savelist()
--- a/bases/computer_lib.py
+++ b/bases/computer_lib.py
@@ -222,7 +222,8 @@ class QuineComputer:
                     alert = f'  {game} is not UTF-8, convert it before loading'
                     print(alert)
                     self.alert_text = GUI.GUIStaticText(self.guiroom, 'txt_screen', alert,
-                        GUI.GUIRect(60, 90, 657, 100), color=GUI.GUIColor(255, 255, 255), bgcolor=GUI.GUIColor(0, 0, 0))
+                        GUI.GUIRect(40, 60, 438, 100, 'pixel', (800,600)),
+                        color=GUI.GUIColor(255, 255, 255), bgcolor=GUI.GUIColor(0, 0, 0))
                     self.alert_text.show()
                     return
                 else:
--- a/bases/utf8_check.py
+++ b/bases/utf8_check.py
@@ -1,0 +1,12 @@
+import VS
+import os
+
+def is_utf8(save_game):
+    try:
+        file_path = VS.getSaveDir() + os.path.sep + save_game
+        with open(file_path, encoding='utf-8') as in_file:
+            text = in_file.read()
+        return True
+    except Exception as ex:
+        print(ex)		
+        return False


bases/computer_lib.py
```diff
@@ -216,7 +217,16 @@ def change_text(self, button_index):
                 ShowProgress.activateProgressScreen('loading',3)
                 import dj_lib
                 dj_lib.enable()
-                VS.loadGame(self.picker_screen.items[self.picker_screen.selection].data)
+                game = self.picker_screen.items[self.picker_screen.selection].data
+                if not is_utf8(game):
+                    alert = f'  {game} is not UTF-8, convert it before loading'
+                    print(alert)
+                    self.alert_text = GUI.GUIStaticText(self.guiroom, 'txt_screen', alert,
+                        GUI.GUIRect(60, 90, 657, 100), color=GUI.GUIColor(255, 255, 255), bgcolor=GUI.GUIColor(0, 0, 0))
```
stephengtuggy, line 225: We'll need to test this with different game resolutions / window sizes, to make sure it works as desired

stephengtuggy (commented):
Overall, looks good so far.

bases/computer_lib.py
P-D-E: Indeed; for the record, these values are based on a 1600x900 window

BenjamenMeyer: yes, if we could make this dynamic for placement and test at various resolutions / window sizes then this should be good. We'll have to make ports to PWCU (for @ermo ).

P-D-E: With these values,
```
GUI.GUIRect(40, 60, 438, 100, 'pixel', (800,600))
```
the alert looks good on my current 1600x900 window and, if I resize the window dynamically (or maximize it), it gets repositioned along with the whole background. It works reasonably well even if I shrink the window quite a lot.

stephengtuggy (approved):
Looks good to me, although I haven't play tested it yet.

ermo:
@P-D-E

Can you help me understand why the CPP check is also necessary?

ermo:
@P-D-E

Can you help me understand why the C++ check is also necessary?

P-D-E:
@ermo Sure, this one only covers the case of game loaded from the main menu; when you're docked in-game and load a game from the station/ship computer, that's handled all inside C++.

Loki1950 (approved):

ermo:
> @ermo Sure, this one only covers the case of game loaded from the main menu; when you're docked in-game and load a game from the station/ship computer, that's handled all inside C++.
> [This](https://github.com/P-D-E/VS-PoC/blob/main/utf8-check-in-game.png) is what it looks like.

Then isn't it wiser to implement a pure C++ UTF-8 check (which I believe you did already) and export that to the VS python module, such that both cases exercise the same utf8 check code and yield the same results and thus minimize the necessary python checks? 

Having two different check-if-UTF-8 functions seems backwards?

P-D-E:
@ermo You sure have a point; exporting the C++ check to the VS python module can of course be done but I was trying to touch the engine the least I could. I hoped for "not at all" but as we found out, that's not the case.

The main reasons for the Py check are/were:
- I got there quick while the C++ side was still unclear to us and we were figuring out how and where to solve it.
- It's easy to test and tune without compilation.
- It's a lot simpler than the C++ check.
- It faces the issue directly and doesn't require altering the VS python module.
- It works by itself and doesn't depend on a PR on a different repository (same as above but a different PoV).
- Modifications to the asset code are needed anyway, to call which-ever check we choose.

That said, your point is just as legit, so you all, what do you think?
Does this go as is or should we make it use the C++ check?

BenjamenMeyer:
@ermo the pure Python check here is the most reliable we have. The C++ side will require more parsing of string data to detect invalid characters.

That said, I'd be great if we could provide this kind of Python code with the engine itself, but we're probably a long way from that since the engine doesn't currently distribute Python code...it really should, but that's another discussion.

stephengtuggy:
I think I agree that we should go ahead and merge this PR as is. I don't think we're ready yet to alter the VS python module (or in other words, what the Vega Strike C++ code exposes to Python). That can come at a later date.

Loki1950:
The Python stub embedded in the C++ allows it to parse the Py scripts in our assets.

stephengtuggy:
What do you think, guys? Can we go ahead and merge this, or is further play testing required?

stephengtuggy:
> @ermo the pure Python check here is the most reliable we have. The C++ side will require more parsing of string data to detect invalid characters.
> 
> That said, I'd be great if we could provide this kind of Python code with the engine itself, but we're probably a long way from that since the engine doesn't currently distribute Python code...it really should, but that's another discussion.

I think I've seen a small amount of Python code on the engine side.

ermo:
> > @ermo the pure Python check here is the most reliable we have. The C++ side will require more parsing of string data to detect invalid characters.
> > 
> > That said, I'd be great if we could provide this kind of Python code with the engine itself, but we're probably a long way from that since the engine doesn't currently distribute Python code...it really should, but that's another discussion.
> 
> I think I've seen a small amount of Python code on the engine side.

If we could do it there with @P-D-E 's UTF-8 Python code, that'd be awesome and be a simple, single-codepath solution.

Unless I'm misunderstanding, then it "merely" (famous last words!) becomes a question of using the right import in the assets code, yes?

stephengtuggy:
> @stephengtuggy that is the stub I mentioned in my comment.

@Loki1950 Gotcha.

stephengtuggy:
Also, the C++ code and the Python code for loading a game were already different, and functioned differently (in my experience), before this latest set of proposed changes. We will need to sync them up at some point. Probably by having them literally use the same code. But that is a larger project, for a future date, seems to me.

stephengtuggy:
> > > @ermo the pure Python check here is the most reliable we have. The C++ side will require more parsing of string data to detect invalid characters.
> > > That said, I'd be great if we could provide this kind of Python code with the engine itself, but we're probably a long way from that since the engine doesn't currently distribute Python code...it really should, but that's another discussion.
> > 
> > 
> > I think I've seen a small amount of Python code on the engine side.
> 
> If we could do it there with @P-D-E 's UTF-8 Python code, that'd be awesome and be a simple, single-codepath solution.
> 
> Unless I'm misunderstanding, then it "merely" (famous last words!) becomes a question of using the right import in the assets code, yes?
> 
> If it is possible to do it so that C++ can call the UTF-8 Python check from basecomputer.cpp and assets can also call the same UTF-8 Python check (with the proper module import from the engine), I'd be one happy camper and would consider it a sound solution.
> 
> (I hope I'm not misunderstanding the Python C++ stub comment in the present reply?)

So you want to put the UTF-8 Python check in the engine? And have both the assets and the engine C++ code call it?

stephengtuggy:
> > > > > @ermo the pure Python check here is the most reliable we have. The C++ side will require more parsing of string data to detect invalid characters.
> > > > > That said, I'd be great if we could provide this kind of Python code with the engine itself, but we're probably a long way from that since the engine doesn't currently distribute Python code...it really should, but that's another discussion.
> > > > 
> > > > 
> > > > I think I've seen a small amount of Python code on the engine side.
> > > 
> > > 
> > > If we could do it there with @P-D-E 's UTF-8 Python code, that'd be awesome and be a simple, single-codepath solution.
> > > Unless I'm misunderstanding, then it "merely" (famous last words!) becomes a question of using the right import in the assets code, yes?
> > > If it is possible to do it so that C++ can call the UTF-8 Python check from basecomputer.cpp and assets can also call the same UTF-8 Python check (with the proper module import from the engine), I'd be one happy camper and would consider it a sound solution.
> > > (I hope I'm not misunderstanding the Python C++ stub comment in the present reply?)
> > 
> > 
> > So you want to put the UTF-8 Python check in the engine? And have both the assets and the engine C++ code call it?
> 
> If possible then yes.
> 
> It solves the issue the correct way IMHO; single code path, in-engine, assets import from engine and run the check.

Hmm. That might actually be the best approach.

It's Python that is having trouble with the non-UTF-8 characters. We've been trying to check for that condition in C++ code, but it seems to be a lot more work. Why not just have Python code (in the engine) perform the check for invalid characters? And then call that Python code from every place that needs the check.

This might actually end up being the fastest solution, even in the short term. Not sure.

Loki1950:
It will become redundant when only the Py3 version is available as if it is not Utf8 just will not occur at that point the check is really for our legacy users.

ermo:
I'd argue that performance is (somewhat) secondary as this is a problem that only really rears its ugly head during load of older save-games.  I'd also argue that it's much better to have the check and not need it, than need it and not have it!

stephengtuggy:
@ermo I agree re. performance. I meant fastest solution to implement.

@Loki1950 I'm not so sure. I think we may continue to have legacy users migrating to the latest VS for a while yet.

BenjamenMeyer:
Right now we're the VS:UtCS and VS Engine are considered closely tied together - we haven't split them yet so it's okay to do things on the C++ side that the game side is tied to when necessary.

We'll revisit more of that approach later on when we start looking to make them more separate entities; but before we can do that we have to expose some various things like version data from the engine to the game assets so they can validate they can use the engine version.

Loki1950:
The Star Trek mod needed several engine changes as it did not use jump nodes so system to system travel needed extra code.

P-D-E:
I've put up the [C++ draft PR](https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/463) in its current state, with the check currently incorporated in `savegame.ccp / h` and used by `cmd/basecomputer.cpp`, as a starting point.

P-D-E:
Closing here, replaced by https://github.com/vegastrike/Assets-Production/pull/62

PR #62, opened 2021-06-02
Utf8 fgnames

P-D-E:

Please answer the following:

Code Changes:

Issues:
- https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/149

Purpose:
- What is this pull request trying to do?  **Restoring all the 0.5.x flight group name files, converted to utf-8/ascii**
- What release is this for? **0.8.x**
- Is there a project or milestone we should apply this to? **0.8.x**


bases/computer_lib.py
```diff
@@ -216,7 +217,16 @@ def change_text(self, button_index):
                 ShowProgress.activateProgressScreen('loading',3)
                 import dj_lib
                 dj_lib.enable()
-                VS.loadGame(self.picker_screen.items[self.picker_screen.selection].data)
+                game = self.picker_screen.items[self.picker_screen.selection].data
+                if not is_utf8(game):
+                    alert = f'  {game} is not UTF-8, convert it before loading'
+                    print(alert)
+                    self.alert_text = GUI.GUIStaticText(self.guiroom, 'txt_screen', alert,
+                        GUI.GUIRect(60, 90, 657, 100), color=GUI.GUIColor(255, 255, 255), bgcolor=GUI.GUIColor(0, 0, 0))
```
BenjamenMeyer, line 225: This needs testing at different resolutions/window-sizes

universe/fgnames/cities.txt
```diff
@@ -1,3421 +1,3421 @@
...
-Allen Park, MI 
-Allen, TX 
-Allentown, PA 
-Alma, QU., Canada
-Almeria, Spain
-Alpena, MI 
-Altoona, PA 
-Alturas, CA 
-Alva, OK 
+Alingsås, Sweden
```
BenjamenMeyer, line 51: probably should get rid of the excess spaces

BenjamenMeyer (changes requested):
Some minor changes needed, and we'll need to do some play testing with this too

bases/computer_lib.py
P-D-E: Testing should be easy, as the graphic elements (alert included) react dynamically to window resizing, try it yourself. Haven't tested full screen modes, but they should behave the same way.
As for the needed changes, feel free to apply any you see fit, included the updated check call via Director, e.g.
```
import Director
...
if not Director.isUtf8SaveGame(game):
```

universe/fgnames/cities.txt
P-D-E: Wow, there were even tabs, uploading a new version...

bases/computer_lib.py
P-D-E: The updated alert geometry had been lost in revisions, reinstated; also, check replaced with the C++ Director one. Tests welcome!

Loki1950 (approved):

stephengtuggy (commented):
I didn't review the fgnames changes in detail, but the Python code looks fine.

Question: @evertvorster was the one who originally combined these files, I believe. His reasoning was that the contents of most of them are exactly the same. He wanted to deduplicate the files, iin essence.

What do you guys think of that approach? Is there a good reason the files were separate before? Perhaps one that Evert and I are unaware of?

P-D-E:
@stephengtuggy According to `data/modules/fg_util.py` if no faction file is found when needed, the choice falls back to `universe/names.txt` so even removing them all does no practical harm; but there are historical design reasons linking factions to name choices, the Minister of Information might shed some light on that.
As for the exact duplicates, I suppose a link would suffice to preserve variety and faction distinction.

stephengtuggy:
@P-D-E That makes a lot of sense.

@ministerofinformation @evertvorster feel free to chime in here.

stephengtuggy:
I think there's something to be said both for flexibility/mod friendliness and for saving disk space. So yes, maybe links are the way to go. Maybe.

P-D-E:
Possible solution: since `ReadBaseNameList` in `data/modules/fg_util.py` already looks for a few fallbacks,
```
filename = 'universe/fgnames/'+faction+'.txt'
...
f = open (filename,'r')
...
f = open ('../'+filename,'r')
...
f = open ('../universe/names.txt','r')
...
f = open ('universe/names.txt','r')
```
the 3rd attempt could be `f = open ('names.txt','r')` where `names.txt` is only one of the 10 identical files. How about this?

evertvorster:
Real life is getting in the way of me playing with Vega Strike as much as I would have liked. 

Links do not work the same way on Windows, and are not necessary. As long as the common denominator file is the fallback, that would be awesome. 

Adding in a little bit of funky characters in the names do add depth to the experience of the game. This is of course only provided that it is not causing crashes or other issues... stability trumps window dressing every time. 

I will be happy to play test once this one is merged. :)

BenjamenMeyer (approved):

BenjamenMeyer:
We can do links in the future; but I'd suggest we do them in the data schema instead of in the file system.
```
{
    "link": "filename or other identifier",
}
```
Also, if we want lists, we could do a loop'd iteration:
```python
x = ['foo', 'foo1', 'foo2']
for v in x:
    try:
       with open(v, ...) as input_data:
           ...
           break
    except OSError: # parent of IsADirectoryError, FileNotFound, PermissionError, etc.
        continue
```

stephengtuggy (approved):

P-D-E:
We were still discussing file de-duplication, also I second Ben's suggestion about lists instead of the original nested try's, do I open another PR for that?

BenjamenMeyer:
@P-D-E yes please

PR #95, opened 2022-07-13
Convert master part list to json

royfalk:
Provide a master_part_list.json file for the engine to use.
Also provide a script - migrate_parts.
Once the script is used, you need to manually delete the headers and add 0 in empty price fields.

See also https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/711
Code Changes:

Issues:
- None identified

### DIFF
--- a/units/migrate_parts.py
+++ b/units/migrate_parts.py
@@ -1,0 +1,35 @@
+import json
+
+keys = ('file','categoryname','price','mass','volume',	'description')
+
+def parseLine(line):
+    values = line.split(',')
+    dictionary = {}
+    i = 0
+    
+    key = values[0]
+    if len(key) == 0:
+        return
+    
+    for value in values:
+        value = value.replace("~", ",")
+        dictionary[keys[i]] = value
+        i = i + 1
+
+    master_array.append(dictionary)
+
+master_array = []
+
+with open("../master_part_list_2.csv", "r") as parts_file:
+    lines = parts_file.readlines()
+    first_line = True
+    for line in lines:
+        if first_line:
+            first_line = False
+            continue
+        line = line.rstrip('\n')
+        parseLine(line)
+
+json_object = json.dumps(master_array, indent = 4)
+with open("../master_part_list.json", "w") as json_file:
+    json_file.write(json_object)
--- a/units/migrate_parts.py
+++ b/units/migrate_parts.py
@@ -1,9 +1,11 @@
 import json
 import sys
 
-master_array = []
+
 
 def main():
+    master_array = []
+    
     with open("../master_part_list_2.csv", "r") as parts_file:
         lines = parts_file.readlines()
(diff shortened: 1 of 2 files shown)


Loki1950 (approved):
Take your time Roy get it right the first time ;)

BenjamenMeyer (changes requested):
like the direction; one minor tweak to how the two functions interact in this form

units/migrate_parts.py
```diff
@@ -0,0 +1,43 @@
...
+            if first_line:
+                first_line = False
+                continue
+            line = line.rstrip('\n')
+            parseLine(line)
+
+    json_object = json.dumps(master_array, indent = 4)
+    print(json_object)
+    with open("../master_part_list.json", "w") as json_file:
+        json_file.write(json_object)
```
BenjamenMeyer, line 20: instead of separately dumping on line 17 and then writing it on line 20, you could do:
```python
with open(...) as json_file:
    json_file.dump(master_array, indent=4)
```

You could even technically make this more efficient by converting the earlier data parsing into a generator where `parseLine` is yielded allowing processing a really huge files (not that we'd have files that large):
```python
def parseFile(parts_file):
   first_line = parts_file.readline() # drop the first line
   for line in parts_file.readlines(): # everything else
       yield parseLine(line.rstrip('\n'))

with open("../master_part_list.json", "w") as json_file:
    with open("../master_part_list_2.csv", "r") as parts_file:
        json_file.dump(parseFile(parts_file), indent=4)
```

royfalk: There's really no need to go that far. It's a one off script for a specific file. It doesn't get run every time by the game.

stephengtuggy (approved):
LGTM

units/migrate_parts.py
```diff
@@ -0,0 +1,44 @@
...
+        first_line = True
+        for line in lines:
+            if first_line:
+                first_line = False
+                continue
+            line = line.rstrip('\n')
+            master_array.append(parseLine(line))
+
+    with open("../master_part_list.json", "w") as json_file:
+        json.dump(master_array, json_file, indent=4)
```
BenjamenMeyer, line 20: would be slightly faster and more memory efficient with the generator/comprehension methods I showed, but good enough - not going to block on that

BenjamenMeyer (approved):

units/migrate_parts.py
BenjamenMeyer: true - if the file was super large (GB in size) then you'd have to just to keep it within memory boundaries.
As noted, with the change it's good enough. Thanks!

units/migrate_parts.py
royfalk: This is a one-off conversion script for a very specific file. This is the wrong hill to optimize on.

PR #96, opened 2022-07-30
CMakeLists.txt: Add file master_part_list.json

stephengtuggy:

Please answer the following:

Code Changes:
- [X] This is an installer system change only

Issues:
- Fixes https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/720

Purpose:
- What is this pull request trying to do? Add a missing file to the list of files to install
- What release is this for? 0.9.x
- Is there a project or milestone we should apply this to? 0.9.x

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -123,6 +123,7 @@ INSTALL(FILES Version.txt DESTINATION share/vegastrike)
 INSTALL(FILES factions.xml DESTINATION share/vegastrike)
 
 INSTALL(FILES master_part_list.csv DESTINATION share/vegastrike)
+INSTALL(FILES master_part_list.json DESTINATION share/vegastrike)
 
 INSTALL(FILES py27_modules.7z DESTINATION share/vegastrike)
 


royfalk:
I admit this makes me uneasy. Here's a file I never knew existed and a crash I didn't catch.
Did I not see it because of cmake cache? 
I'd appreciate if you'd write a few lines about it and maybe update the PR validation guide.
Anyway, LGTM.

royfalk (approved):
LGTM.

evertvorster:
I can only guess here, but as I only point vegastrike at the directory that is made by the install part of CMakeList I would catch errors like these. My question now is, how are the other packages not affected? Surely they only install what is specified in Cmake?

BenjamenMeyer:
> I can only guess here, but as I only point vegastrike at the directory that is made by the install part of CMakeList I would catch errors like these. My question now is, how are the other packages not affected? Surely they only install what is specified in Cmake?

We've just started changing things in VS Engine master to use the JSON by default; other stuff still relies on the CSV. We should be installing the JSON too.

BenjamenMeyer (approved):

PR #97, opened 2022-09-05
Fix units.json

royfalk:
Prepare units.csv for parsing by replacing commas with pipes (not included in commit).
Simplify parser.py by removing quotes and converting pipes back to commas.
Fix the units.json file which had errors.

Someone will need to look at Privateer. I don't plan to deprecate CSV format, but it will be parsed with new code. I also noticed a whole bunch of unused keys/attributes such as 'hidden hold'. These may be used by the privateer side.

Please answer the following:

Code Changes:
- [ ] Have the PR Validation Tests been run? In progress, pending work on  the second PR on the engine side.

Issues:
- Once the engine started using the json, some bugs in the parser popped up.


Loki1950:
The hidden hold is used in privateer for contraband in a few missions in the story line.

units/units.json
```diff
@@ -41747,4 +41747,4 @@
         "Moment_Of_Inertia": "0",
         "Mounts": "{ShieldBreaker;;;;;;;;;;;;;;;1;1}"
     }
-]
+]
```
stephengtuggy, line 41750: The only difference I see in units.json is that it no longer has the blank line at end of file. Is this intentional?

stephengtuggy (commented):
Looks good overall. Just one question

Loki1950 (changes requested):
Same comment as @stephengtuggy

units/units.json
royfalk: I had a whole elaborate answer about a bug I fixed. However, sometime between the previous commit and this one, I made changes to parser.py which caused it to bork the json file.
Which, this commit fixes...

I'll have to consider whether to close the PR or just remove the JSON. Something to do tomorrow.

stephengtuggy: Sounds good. I think there are other valid changes in this PR, so I would leave it open and just remove the JSON file.

royfalk: units.json now has several real changes. I merged units.description into it as well as fixed a few units that were missing attributes (but don't remember which unfortunately.)
I did not intend to reuse this PR. I simply forgot it existed and kept working on it concurrently with the engine code.

stephengtuggy (approved):
LGTM

Loki1950 (approved):

Issue #107, opened 2024-03-28
module "imp" was deprecated in python 3.12


david-geiger:
The module "imp" should be replaced by 'importlib" as it was deprecated in python3.12.

This proposed patch fixes this:

[0001-Migrate-from-imp-to-importlib.patch.txt](https://github.com/vegastrike/Assets-Production/files/14784754/0001-Migrate-from-imp-to-importlib.patch.txt)

BenjamenMeyer:
@david-geiger feel free to submit a PR; otherwise I'll file it when I get a chance. I'd rather see you get the credit though.

david-geiger:
Ok, I'll do it :)

stephengtuggy:
Fixed as of #140

PR #109, opened 2024-04-21
Migrate from imp to importlib

david-geiger:
- the 'imp' module is deprecated from years and now Python 3.12 finally drops completely support of 'imp' module.
- fix for #107


stephengtuggy:
#140 takes care of these fixes for 0.9.x. Closing in favor of #140

PR #110, opened 2024-04-21
Migrate from imp to importlib

david-geiger:
- the 'imp' module is deprecated from years and now Python 3.12 finally drops completely support of 'imp' module.
- fix for #107

### DIFF
--- a/modules/custom.py
+++ b/modules/custom.py
@@ -1,7 +1,7 @@
 import VS
 import traceback
 import sys
-import imp
+import importlib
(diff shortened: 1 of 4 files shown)


BenjamenMeyer (approved):
Can be merged once play tested

stephengtuggy (approved):
I made these same changes today, independently, and tested them on macOS 11. The code works with these changes -- in fact, they may fix some broken functionality.

PR #130, opened 2024-10-27
A python script to generate ship view

royfalk:
This PR starts moving a lot of the ship view and later similar code from base_computer.cpp to python. The script can be run independently and pull data from units.json/llama.begin. 

This works with https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/899.

Please answer the following:

Code Changes:
- [x] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation. 

Issues:
- This will break when we merge a PR that saves a resource in the form x/y/z. We'll need to modify this further to support.

### DIFF
--- a/python/base_computer/ship_view.py
+++ b/python/base_computer/ship_view.py
@@ -1,0 +1,304 @@
+import locale
+import math
+import json
+
+
+newline = "#n#"
+red = "#c1:.3:.3#"
+green = "#c0:1:.5#"
+light_grey = "#c.75:.9:1#"
+grey = "#c.6:.7:.8#"
+light_yellow = "#c.675:.925:.825#"
+end_color = "#-c"
+
+non_combat_speed_multiplier = 1
+megajoules_multiplier = 1
+
+llama_begin = {}
+
+with open('config.json', 'r') as file:
+    data = json.load(file)
+    non_combat_speed_multiplier = data['components']['drive']['non_combat_mode_multiplier']
+    megajoules_multiplier = data['constants']['megajoules_multiplier']
+    
+with open('units/units.json', 'r') as file:
+    data = json.load(file)
+    
+    for ship in data:
+        if ship['Key'] == 'Llama.begin':
+            llama_begin = ship
+            break
+
+
+# Format large number
+def lnum(ship_stats, key, divider = 1.0):
+    num = float(ship_stats[key])/divider
+    num = '{:n}'.format(num)
+    return num
+
+def get(ship_stats, key):
+    return ship_stats[key]
+
+def get_bool(ship_stats, key):
+    s = ship_stats[key]
+    if(s == 'TRUE' or s=='1'): 
+        return True
+    return False
+    
+def get_int(ship_stats, key):
+    return int(ship_stats[key])
+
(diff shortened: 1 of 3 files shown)


evertvorster:
Hi there!

Before building this, I see that you have to merge in the change where the config.json is installed into this branch, and so resolve that conflict. 
Also, there is another new file that also is not installed through the CMakeLists.txt. 

It is a bit tedious, but makes things easier down the line. 

I don't see how the game engine is taking advantage of this new python script. It that to be implemented at a later stage when this is in master?

python/base_computer/ship_view.py
```diff
@@ -0,0 +1,304 @@
...
+
+non_combat_speed_multiplier = 1
+megajoules_multiplier = 1
+
+llama_begin = {}
+
+with open('config.json', 'r') as file:
+    data = json.load(file)
+    non_combat_speed_multiplier = data['components']['drive']['non_combat_mode_multiplier']
+    megajoules_multiplier = data['constants']['megajoules_multiplier']
```
BenjamenMeyer, line 22: since `config.json` would have to be in a specific place we need to either:
(a) define a working directory for the active "game"
(b) expose these as Python Objects instead of making each script load it if it wants access. This would eliminate the `open` and `json.load` steps; and we'd have to define the interface - either as a pure Python Dictionary/Object or as something else. A first step might be just having a `load` method that would load the contents and return the `data` object like you have above.

I'd rather `b` between the two.

python/base_computer/ship_view.py
```diff
@@ -0,0 +1,304 @@
...
+    
+with open('units/units.json', 'r') as file:
```
BenjamenMeyer, line 24: same here

python/base_computer/ship_view.py
```diff
@@ -0,0 +1,304 @@
...
+    text += get_durability(ship_stats)
+                      
+    return text
+
+# Test function
+# Useful to check python script for correctness before running VS
+
+t = get_ship_description(llama_begin)
+t = t.replace('#n#','\n')
+print(t)
```
BenjamenMeyer, line 302: if this is for testing, then it probably should be wrapped. Usually if its a test to be called on the commandline (e.g just see the script output) then using:
```python
if __name__ == "__main__":
   # python block
```
would be appropriate. If we can control the name and calling it in a test, then may be we call it `__vegastrike_test_main__`  instead...but I doubt that'd we'd have that control.

BenjamenMeyer (approved):
Approving to get it in; but I'd like to see the File `open` and `json.load` move to the C++/engine side so we can present a more consistent interface to the game; unless of course we're expecting that `config.json` is a pure Game Asset - in which case we should go with option `a` in the comment and expose a good method for knowing that from the VS side - but all that can be evolution of this PR.

python/base_computer/ship_view.py
BenjamenMeyer: Thinking more about this...option `a` might actually be a better solution for game asset data. But let's evolve the conversation on this beyond just this PR.

BenjamenMeyer (approved):

PR #129, opened 2024-10-25
Complementing changes to engine 'Implement Drive and Afterburner'

royfalk:
This complements https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/896
- Add support for AfterburnerUpgrade
- Modify Llama.begin with reactor03 and correct SPEC capacitor. This will allow SPEC with game start. 
- Remove Llama.begin from units.json.

Thank you for submitting a pull request and becoming a contributor to Vega Strike: Upon the Coldest Sea.

Please answer the following:

Code Changes:
- [x] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

PR #134, opened 2024-11-20
Fix issue caused by new drive PR #896

royfalk:
Resource now save in format x/y/z which breaks ship_info

Code Changes:
- [X] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation minimal. check issue is fixed.


evertvorster:
Testing against current master of the Engine. (If this is the wrong branch, some indication as to which branch of the engine code to test against would be appreciated)
When starting a new campaign, in ship info shield info indicates that there is no shield installed. 
In the upgrade bay, I can see that there is, in fact, a shield installed. 
When purchasing refuel and repair, I get a message that the ship has no damage, and that there would be no charge. Normally there is some work done, and the ship maximum speed changes from 120 to 125. 
When launching this ship, there is no shielding on the ship.

royfalk:
> Testing against current master of the Engine. (If this is the wrong branch, some indication as to which branch of the engine code to test against would be appreciated) 

No need to test more than ship view.
Not a bad idea to test against master.
But really, you should test against PR #896, which is task_real_drive_comp.

evertvorster:
Hi there!
I just tested this branch against the task_real_drive_comp branch of the engine. 
Unfortunately this same issue remains. 
When starting a new campaign, the shields of the Llama.begin does not seem to be interpreted correctly. 
It shows no shielding information in ships info, even though a shield is present in the ship upgrade window. Also, when launching, there is no shield on the ship.
For some reason Oswald's ship also seems damaged, not quite sure if this is related.

royfalk:
It is. Are you by any chance using the new units.json from https://github.com/vegastrike/Assets-Production/pull/133 ?
That would be consistent with it.

evertvorster:
Hi there!
Is there any way I can check? As far as I know, the Arch build system totally removes all the files from the previous install before installing the current branch. 
However, I will test again with this branch.

evertvorster:
Hi there!
I think I found the issue, and it was on my side. I was under the impression that telling git to checkout a certain branch would show only the files from that branch. 
Anyways, I totally removed the source tree, and checked out this branch. 
Now that issue with the shields is gone. I'll give it some more testing and see if I can find any show stoppers. 
Thanks for your patience.

royfalk:
Merged changes from #129 .

royfalk:
@evertvorster is this OK? I already merged https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/896 and then discovered the matching Assets is still not approved. You can test this branch against engine/master.

evertvorster (approved):
Yes, this is the branch I tested against, and should be merged.

PR #136, opened 2024-11-25
Finish moving content to ship view on python

royfalk:
engine: https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/917

Please answer the following:

Code Changes:
- [x] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation Only on base_computer


evertvorster (changes requested):
Hi there!

Not sure if this error is in Assets or Engine, but the combination of these two branches, I see that orthogonal accelleration for all ships show as 0. 

When loading a saved file, the orthogonal accelleration are normal values. 

When starting a new campaign, the orthogonal accelleration is also 0 on the Llama. 
This has the effect that if you accelerate in a straight line, and then turn a little, your speed never drops back down to 0. 

I'm assuming that the orthogonal acceleration is not read properly, or somehow got corrupted.

royfalk:
It's https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/909.

BenjamenMeyer (approved):

PR #137, opened 2024-11-26
Task more config

royfalk:
To test, play with components|fuel|factor. Set it to 0.1 to see it drop really fast. Something around 100-2000 is right. 
Same for the other factors. 

Storage factors, fuel, energy and ftl_energy impact how much is stored in the capacitor/fuel.
You should be able to see it in ship view.

For consumers, factors affect consumption. Play with ftl_drive|factor (=100) to see how it eats up ftl_energy.

**A note about naming:**
I moved away from warp/spec to a neutral ftl_* to make it:
a. consistent
b. game neutral


royfalk:
OK. Please attach vegastrike.config, config.json and the command line you use to run the game.
e.g. ./vegastrike-engine -d../data ...

evertvorster:
Sure thing:
[vegastrike.config.zip](https://github.com/user-attachments/files/17920592/vegastrike.config.zip)

[config.zip](https://github.com/user-attachments/files/17920600/config.zip)

The "vs" is what is distributed with Assets:
```
[evert@Evert .vegastrike]$ which vs
/usr/bin/vs
[evert@Evert .vegastrike]$ cat /usr/bin/vs
#!/bin/bash
# This is a small wrapper that lets VegaStrike - Upon the Coldest Sea
# run from seemingly anywhere
set -e

VEGASTRIKE_LOCAL_SHARE_DIR="/usr/local/share/vegastrike"
VEGASTRIKE_SHARE_DIR="/usr/share/vegastrike"

#Let's keep a log of the console
if [ -f  ~/.vegastrike/console.log ]; then 
  mv -f ~/.vegastrike/console.log ~/.vegastrike/console.bak
fi
if [ -f ~/.vegastrike/console_err.log ]; then 
  mv -f ~/.vegastrike/console_err.log ~/.vegastrike/console_err.bak
fi

echo "Starting Vega Strike..."
echo ""

if [ -d "${VEGASTRIKE_LOCAL_SHARE_DIR}" ]; then
    vegastrike -d"${VEGASTRIKE_LOCAL_SHARE_DIR}" "$1"
elif [ -d "${VEGASTRIKE_SHARE_DIR}" ]; then
    vegastrike -d"${VEGASTRIKE_SHARE_DIR}" "$1"
else
    echo "Unknown Game Asset Data Location."
    echo "Do you have the game assets installed?"
    exit 1
fi
[evert@Evert .vegastrike]$
```

evertvorster:
No dot slash, it's installed in /usr/bin, so just "vs"

royfalk:
I updated the engine part. Please try that.

evertvorster:
Hi there!

Using this thread to reply, even though I am pretty sure this is more or an engine thing than an assets thing.

Firstly, with that change applied, Vega Strike now starts up. 

Unfortunately with new ships and with savegames a lot of the ship information is missing, and there is this error message at the top of the ship info screen:
image

royfalk:
Actually, it may well be the engine. 
In the console from which you ran the game, there could/should be a more detailed error message.
Can you please check?

evertvorster:
Hi there. 

Indeed there is:
```
Launching bases for Special/Empty
Adding news
conditioning
nonzeroing
conditioning
nonzeroing
Loading active missions True
Special/Empty  IS NOW AT  Crucible/Cephid_17
Special
Showing splash screen!
Launching bases for Crucible/Cephid_17
Adding news
finding nonloaded quest
quest_tutorial
finding quest
quest_tutorial
nonpfact
Processing News
Traceback (most recent call last):
  File "/usr/share/vegastrike/python/base_computer/ship_view.py", line 25, in <module>
    data = json.load(file)
           ^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/__init__.py", line 293, in load
    return loads(fp.read(),
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
               ^^^^^^^^^^^^^^^^^^^^^^
json.decoder.JSONDecodeError: Invalid control character at: line 54 column 35 (char 791)
Thank you for playing!
```

To get this message, all I did was to start a new campaign. Then I went to the mission computer on Atlantis, and then clicked on ship info.

evertvorster:
Hey there. Mint is a fine distro, and hopefully you are happy with it. 

Even though I am not running it myself at the moment, I was seriously impressed with Neptune OS. 
It's Debian based, with an updated kernel and latest KDE and a good set of default apps. 
If you are not quite happy with Mint, Neptune is a little known gem that is worth a look. 

In any case, I was quite busy the last day as well. It was election day in my home country, and as you might expect from an African country, not everything went smoothly, and the event ended up taking more than a whole day.

royfalk:
I believe I fixed the issue above. Can you check? This is linked to https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/919.

evertvorster:
Hi there. 
That error message seems to be gone, and there is a lot more information there. 
One thing that is strange is the number representation of the fuel capacity. 
Should it not just be an integer?
This is a screenshot from this branch:
image

And here is the current master for comparison:
image

evertvorster:
It's nice to be able to tune all these new features. 
However, it is not quite clear what each of the numbers does. 

What does "non_combat_multiplier" do? Does it set the GCNT:Travel speed? Previously this was hard-coded. 

"fuel" seems to adjust the number for the fuel present on the ship. When adjusting it so that the number is similar to the master branch, the fuel guage seems to empty out pretty quick. Setting it higher makes the number for fuel in the ship go so large that it starts using scientific notation. 

That fuel burn rate still seems awfully high, and I don't see where it can be adjusted. Just adding more fuel to the ship makes for an incongruity because you are adding mass to the ship which is not accounted for in acceleration. 

If I add 4000mt of something dense in the hold of the Llama, the ship is at 13000% of its normal mass, and acts like it. 
Yet, somehow 9000000mt of fuel does not do anything?

I believe the thinking in regards with the fuel was that it is an incredibly dense form of energy, so that not so much of it was used. 

In short, it would be nice to be able to adjust the fuel burn rate as well. 

What does "megajoules_multiplier" affect? Instinctively Mega means a million, or 1E6, this 100 seems odd. 


I read that json does not natively support comments, but having a "_comment" key can have the same effect, and some more information on what each of the new configuration options does would be nice.

royfalk:
> One thing that is strange is the number representation of the fuel capacity.

It's a bug in the save ship function. The same function also "saves" the ship to transfer it to python for ship view. It does not apply the fuel.factor in reverse.

> What does "non_combat_multiplier" do? Does it set the GCNT:Travel speed? Previously this was hard-coded.

Yes

> "fuel" seems to adjust the number for the fuel present on the ship. 

Correct. It was easier to adjust the fuel than to adjust burn rate. It shouldn't matter to you as long as the displayed fuel makes sense.

> That fuel burn rate still seems awfully high, and I don't see where it can be adjusted. 

You can't. It was easier to adjust fuel in one place.

> Just adding more fuel to the ship makes for an incongruity because you are adding mass to the ship 

You're not. Fuel mass is not counted (for now). I just need to adjust the displayed fuel by dividing it by efficiency to get a number. Ideally, we should have a fuel_to_mass factor to actually display something logical, but we're a few weeks away from that.

> What does "megajoules_multiplier" affect? Instinctively Mega means a million, or 1E6, this 100 seems odd.

This is game units to MJ multiplier. So 1.0 in VS is 100MJ in real world. The previous devs were science nerds and tried to ground this in reality.

evertvorster:
> > One thing that is strange is the number representation of the fuel capacity.
> 
> It's a bug in the save ship function. The same function also "saves" the ship to transfer it to python for ship view. It does not apply the fuel.factor in reverse.
This sounds like something that needs to be fixed?

> > Just adding more fuel to the ship makes for an incongruity because you are adding mass to the ship
> 
> You're not. Fuel mass is not counted (for now). I just need to adjust the displayed fuel by dividing it by efficiency to get a number. Ideally, we should have a fuel_to_mass factor to actually display something logical, but we're a few weeks away from that.
What might make more sense for the large numbers in here now would be to change the displayed unit. Rather than call it "metric tonnes" we can call it "kilograms" . From there we can explore adding this weight to the ship if necessary. 

I suppose with a bit of testing we can come up with sane numbers for the various variables, and then adjust them in Assets?

royfalk:
> This sounds like something that needs to be fixed?

Fixed in d076534f51945d685a489101191b461946e857ac on the engine side.

> What might make more sense for the large numbers in here now would be to change the displayed unit. 

This is exactly why moving this to a python script is so great. The cost of playing around with these ideas is much lower. Also, the barrier to entry is much lower. If I didn't say it, you are welcome to play around and contribute to the python scripts. 

> I suppose with a bit of testing we can come up with sane numbers for the various variables, and then adjust them in Assets?

I don't care if game numbers are sane. Just that they are **consistent**. We can then adjust them with a factor in config to whatever we want. Finally, we can adjust them again in the python script to make them "sane" and clear to the player.

Example for fuel - ships can have fuel capacity between 1 and 10. 1 is very short range, fighter to defend stations. 10 is long term exploration. We then adjust the game mechanics with a factor like we actually did for fuel. We can further provide hints to players (divide by half for more realistic and challenging game).
Finally, we can adjust either the original number or the factored one again in shop view to display Metric Tons of Lithium 6 or Kg of Deuterium.

royfalk:
By inconsistent I mean that Llama has a mass of 250 has 25 tons of fuel (1:10) but Ox has a mass of 30mT and fuel of 170kT (1:176). 

Also, 30mT is about x100 what we're currently producing. Assuming consistent scaling, that comes to a ship 50km in length. Even if someone were to build that (does it make economic sense?), the game doesn't do it justice right now. Did you fly alongside an ox and said "this thing doesn't end!" ??

royfalk:
Self approving, as engine branch was approved by both @BenjamenMeyer  and @evertvorster .

PR #138, opened 2024-12-04
Overwrite config.json with player configuration

royfalk:
We are currently in the process of migrating vegastrike.config (old, XML) to config.json (new, JSON).
As a result, the structure of the new file changes regularly and overwrites whatever changes the player has introduced.
As the first thing we ported was the graphics (resolution, monitor), 
this requires the player to manually edit the file after every relevant pull.

Solution
--------
There is a script, conveniently named overwrite_config_defaults, to do this automatically.
Copy all your changes to config.bak, preserving the relevant structure.

Code Changes:

Issues:
- None

Does not fix Engine/issues/921


evertvorster:
Hi there!

Thanks for looking into this. However, this script is still set to operate in a region of the system which is not under the user's control. So, in order to run, it still needs superuser privileges. 

As with vegastrike.config, would it not make more sense to have the engine check for a config.json in the same places and hierarchy as with vegastrike.config?

If a config.json is found in the user's home directory, then just use the config.json in the user's home directory rather than the one in the system. 

This solves a couple of problems:
1. No elevated user privileges are required. 
2. Since it is not part of the Assets package, it does not get overwritten when Assets are updated.

If you want to be fancy about it, you can put a version number inside the config.json, so that when the items in the config.json changes it would be possible to warn the user that the config file is of an older version and needs to be merged with the newer version in /usr/share/vegastrike.

royfalk:
OK. This is somewhat more complicated than I bargained for.
I believe the issue is fixed for people with sudo (specifically you).
I've detached this PR from the issue. 

If this solves the issue for you, let's merge this and revisit the issue again in a few weeks.

evertvorster:
Hmm, for me it is easier to just have a copy of the config.json somewhere else and just sudo cp it back into position. 
Maybe it is better to just forget about this PR and focus on something else. 
We can revisit this issue when it becomes an issue for more people.

royfalk:
> Hmm, for me it is easier to just have a copy of the config.json somewhere else and just sudo cp it back into position. 

It is now. But in a month or two, you'll have multiple changes to a much larger file. 

Don't forget, config.json is supposed to grow rapidly in size.

evertvorster:
For now, I'll just copy over the file manually. If it changes significantly, I'll copy the new one and make the changes needed in my copy. 
At some point in the future the game engine should save the config file in a destination which is not a system directory, and prefer that copy over the system one. However, we can cross that bridge when we get to it.

python/utilities/overwrite_config_defaults.py
```diff
@@ -0,0 +1,33 @@
...
+
+def recursive_overwrite(config_json, config_bak):
+    if isinstance(config_bak, dict) or isinstance(config_bak, list):
+        for key in config_bak:
+            # The check is repeated because if we pass a leaf to recursive_overwrite
+            # it is passed by value
+            if isinstance(config_bak[key], dict) or isinstance(config_bak[key], list):
+                recursive_overwrite(config_json[key], config_bak[key])
+            else:
+                config_json[key] = config_bak[key]
```
BenjamenMeyer, line 14: might be able to use a Python dict merge here:
```
dict_a = {}
dict_b = {}
a.update(b)
```

royfalk:
The engine now overwrites the assets config.json with the user folder config.json.
This makes this PR redundant.

PR #139, opened 2024-12-08
Recreate degraded speed governor in llama.begin

evertvorster:

Please answer the following:

Code Changes:

Issues:
- With the new drive implementation in the engine, the beginning llama was stuck at a speed of 120, even after a repair and refuel which would previously bump the speed to be the same as all the other Llamas that you can get in VegaStrike. 

Purpose:
- What is this pull request trying to do? - Recreate historical game play. In effect, the beginning Llama has some issues that sets the speed governor to 120. On the first refuel and repair, these issues are resolved, and then it should have a speed governor setting of 125.
- What release is this for? - Git master. 
- Is there a project or milestone we should apply this to? - Nope.

Needed to remove the Afterburner_Speed_Governor from llama.begin too, because otherwise the ship would forever be stuck with a less performant overdrive as compared to all the other llamas in the game. With the line removed, overdrive governor gets the default 125 value, which is in line with what the llama.begin would get with the first repair and refuel. 
It is highly unlikely that you would actually get to a place where you would purchase an overdrive without first having purchased a repair and refuel. Even if you did, the effect on gameplay would be far less than having a less performant overdrive after a repair and refuel.


royfalk (approved):
Please open an issue in engine - ships do not have the option of no afterburner.
Approved.

PR #144, opened 2025-01-15
ship_view.py: Fix a few things, best I can figure out

stephengtuggy:
Code Changes:
- [X] (Mostly) Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues:
- I don't think a ticket has been filed yet for the errors this PR addresses

Purpose:
- What is this pull request trying to do? Fix some errors being thrown in ship_view.py

### DIFF
--- a/python/base_computer/ship_view.py
+++ b/python/base_computer/ship_view.py
@@ -38,9 +38,22 @@ def get_unit(key):
         return None
 
 # Format large number
-def lnum(ship_stats, key, divider = 1.0):
-    num = float(ship_stats[key])/divider
-    num = '{:n}'.format(num)
+def lnum(ship_stats, key, divider = 1.0) -> str:
+    try:
+        num = float(ship_stats[key])/divider
+        num = '{:n}'.format(num)
+    except KeyError:
+        print('KeyError in lnum: ', key, ' not found in ship_stats')
+        return ''
+    except ZeroDivisionError:
+        print('ZeroDivisionError in lnum')
+        return ''
+    except TypeError:
+        print('TypeError in lnum')
+        return ''
+    except ValueError:
+        print('ValueError in lnum')
+        return ''
     return num
 
 def get(ship_stats, key):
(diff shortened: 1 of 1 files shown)


evertvorster (approved):
Looks really good. 
Please merge at the same time as the other fixes some stuff PR in engine.

python/base_computer/ship_view.py
```diff
@@ -56,9 +69,18 @@ def get_int(ship_stats, key):
     return int(ship_stats[key])
 
 def get_dbl(ship_stats, key, divider = 1.0):
-    fl = float(ship_stats[key])/divider
-    #s = f"{key} {fl}"
-    #print(s)
+    try:
+        fl = float(ship_stats[key])/divider
+    except KeyError:
+        print('KeyError in get_dbl: ', key, ' not found in ship_stats')
```
BenjamenMeyer, line 75: FYI - since we're using Py3 we can use[ f-Strings](https://docs.python.org/3/tutorial/inputoutput.html#tut-f-strings) here we makes formatting really nice and easy:
```python
print(f'KeyError in get_dbl: {key} not found in ship_stats')
```

BenjamenMeyer (approved):
LGTM - left a comment on something that we could do a little better now that we're using Py3 - probably a little more efficient than the older methods too.

python/base_computer/ship_view.py
royfalk: Should it be f"blurb" and not f'blurb'?

royfalk (approved):
LGTM

python/base_computer/ship_view.py
BenjamenMeyer: @royfalk it doesn't really matter which; typically you use one if you want to use the other:
```python
f"blurb 'b' "
f'blurb "b" '
```
they both denote a string.

PR #151, opened 2025-01-23
Windows-Release.yml: invoke CPack via CMake:… (0.10.x)

stephengtuggy:
… `cmake --build . --target package`

Code Changes:
- [X] This is a CI/build system change only (and I have run the cmake command locally to make sure that it works)

Issues:
- N/A

Purpose:
- What is this pull request trying to do? Fix the Windows packaging of the vsUTCS Assets, for 0.10.x and following

### DIFF
--- a/.github/workflows/Windows-Release.yml
+++ b/.github/workflows/Windows-Release.yml
@@ -56,7 +56,7 @@ jobs:
         run: script/build
 
       - name: Package it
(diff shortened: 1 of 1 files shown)


BenjamenMeyer (approved):
might have to adjust with the cmake refactor on master/0.10.x

PR #156, opened 2025-02-07
Support engine change to Radar component

royfalk:
Support for https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/804

- Fix too large radar range.
- Damage Llama.begin, both for testing and for gameplay in general. We now support damaged components in an elegant way and there's no reason a refurbished, second-hand Llama wouldn't have a lot of dings to fix.

Code Changes:
- [x] Have the PR Validation Tests been run? 

Issues:
- Radar range for three units was too large 3x10^9 instead of 10^8. This somehow caused Resource to crash. Not sure why. Double definitely supports more than this number. In any case, this is definitely a type, as all other units are 3x10^8, so fixed it and moved on.


stephengtuggy (approved):
Looks good to me.

BenjamenMeyer (approved):

PR #160, opened 2025-02-18
Fix issue in upgrade view

royfalk:
where radar range caused script to crash.

This PR is only for this issue, as previous PR was already merged.

Please answer the following:

Code Changes:

### DIFF
--- a/python/base_computer/upgrade_view.py
+++ b/python/base_computer/upgrade_view.py
@@ -150,7 +150,27 @@ def process_resource(key, resource_text):
     original = float(parts[original_index])
 
     return f"{format_number(damaged)} (orig: {format_number(original)})"
+
+def process_range(text):
+    parts = text.split('/')
+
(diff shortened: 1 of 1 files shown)


python/base_computer/upgrade_view.py
```diff
@@ -150,7 +150,27 @@ def process_resource(key, resource_text):
...
+    original_index = 0
+
+    if len(parts) == 2:
+        damaged_index = 0
+        original_index = 1
+    elif len(parts) == 3:
+        damaged_index = 1
+        original_index = 2
+    else:
+        return text
```
BenjamenMeyer, line 167: do we want to log the last case to help with debugging? Or would it always be valid?
Some comments on the above would be helpful to understand the 3 cases.

royfalk: Added documentation.

PR #164, opened 2025-03-09
Remove useless function `ShowProgress.setProgressBar()`

vincele:
The C++ engine does not do anything with it.
See: https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/3517e33b96b4ce9e0214c736d7379aa8a2b39d79/engine/src/universe_util.cpp#L204

When/if this is merged, I have the part removing the (then also unused) C++ code.

Code Changes:
- [X] Have the PR Validation Tests been run? compiled engine and run with this PR, on linux/musl

Issues:
- Not sure if this should go to Assets Master, even if I've read [the issue 14](https://github.com/vegastrike/Assets-Masters/issues/14) and [the issue 4](https://github.com/vegastrike/Assets-Production/issues/4). As I've done that here for testing purpose, I created the PR here. I'll move upon request / clarification.

Purpose:
- What is this pull request trying to do? Remove unused code
- What release is this for? Next
- Is there a project or milestone we should apply this to? Next

### DIFF
--- a/modules/ShowProgress.py
+++ b/modules/ShowProgress.py
@@ -53,8 +53,3 @@ def setProgressMessage(screen_id,message):
         if prefix != "":
             prefix += ": "
         VS.showSplashMessage(prefix+message)
-
-def setProgressBar(screen_id,progress):
-    global __showing
-    if screen_id == __showing:
-        VS.showSplashProgress(progress)
--- a/modules/fg_util.py
+++ b/modules/fg_util.py
@@ -397,7 +397,6 @@ def DeleteLegacyFGLeftovers():
         count += 1
         pct = count*100/len(allsys)
         if pct != oldpct:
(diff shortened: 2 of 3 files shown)


BenjamenMeyer:
This is the correct place to file a PR as these are the assets we actually use. I don't know that any of us know how to build stuff from Assets Master right now which is closer to the raw data; but it doesn't produce a game - only Assets Production does.

We file the issues against Assets Master b/c its more central. But we do need to figure out Assets Master better too.

BenjamenMeyer:
Perhaps the better answer would be for us to figure out how to implement the Progress Bar in the engine which will give better feedback on data loading in the game.
@royfalk what are your thoughts?

royfalk:
This is definitely one of those places where finishing the work the original devs started is possible and easy to do. I vote to keep it.

PR #166, opened 2025-03-22
Fix travel mode multiplier

stephengtuggy:
Code Changes:

Issues Addressed:
- Addresses https://github.com/vegastrike/Assets-Masters/issues/63

Known Issues Remaining:
- The travel mode multiplier isn't fixed. Still stuck at 100x most of the time. More troubleshooting is needed.

Purpose:
- What is this pull request trying to do? Fix https://github.com/vegastrike/Assets-Masters/issues/63 

### DIFF
--- a/config.json
+++ b/config.json
@@ -2,8 +2,8 @@
 "general": {},
 
 "graphics": {
-    "resolution_x": 2560,
-    "resolution_y": 1600,
+    "resolution_x": 1920,
+    "resolution_y": 1080,
     "screen": 0
 },
   
--- a/config.json
+++ b/config.json
@@ -6,40 +6,40 @@
     "resolution_y": 1080,
     "screen": 0
 },
-  
+
 "components": {
     "afterburner": {
     },
-        
+
     "drive": {
         "non_combat_mode_multiplier": 1000
     },
-    
+
     "energy": {
         "factor": 1
     },
-    
+
     "fuel": {
         "fuel_equals_warp": false,
         "factor": 600
     },
-    
+
     "ftl_drive": {
         "factor": 0.1
     },
-    
+
     "ftl_energy": {
         "factor": 1.0
     },
-    
+
     "jump_drive": {
         "factor": 1.0
     },
-    
+
     "reactor": {
         "factor": 1.0
     }
-    
+
 },
 
 "constants": {
@@ -53,6 +53,6 @@
     "introduction": 
         "Welcome to Vega Strike!\nUse #8080FFTab#000000 to afterburn (#8080FF+,-#000000 cruise control),\n#8080FFarrows#000000 to steer.\nThe #8080FFt#000000 key targets objects; #8080FFspace#000000 fires at them & #8080FFa#000000 activates the SPEC drive.\nTo go to another star system, buy a jump drive for about 10000 credits,\nfly to a wireframe jump-point and press #8080FFj#000000 to warp to a near star.\nTarget a base or planet;\nWhen you get close a green box will appear. Inside the box, #8080FFd#000000 will land."    
 },
-  
+
 "advanced": {}
 }
--- a/config.json
+++ b/config.json
@@ -12,7 +12,7 @@
     },
         
     "drive": {
-        "non_combat_mode_multiplier": 100
+        "non_combat_mode_multiplier": 1000
     },
     
     "energy": {
--- a/config.json
+++ b/config.json
@@ -2,8 +2,8 @@
 "general": {},
 
 "graphics": {
-    "resolution_x": 1920,
-    "resolution_y": 1080,
+    "resolution_x": 2560,
+    "resolution_y": 1600,
     "screen": 0
 },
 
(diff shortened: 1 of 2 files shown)


vincele (approved):
LGTM

stephengtuggy:
@vincele Thanks. Unfortunately, this doesn't seem to fix the problem completely. I will need to do some further troubleshooting.

royfalk:
Can we check that the old code was x1000?
I'm AFK for a few more days.

stephengtuggy:
I'm positive that the old code was x1000. I remember it distinctly.

I just added one more, one-liner change to this PR, to fix the FTL Capacitor 2 issue. Hopefully that's all right.

royfalk:
> I'm positive that the old code was x1000. I remember it distinctly.

You were right.
From computer.cpp in 0.9.x:

```
float Computer::max_speed() const {
    static float
            combat_mode_mult = XMLSupport::parse_float(vs_config->getVariable("physics", "combat_speed_boost", "100"));
    return (!combat_mode) ? combat_mode_mult * max_combat_speed : max_combat_speed;
}

float Computer::max_ab_speed() const {
    static float
            combat_mode_mult = XMLSupport::parse_float(vs_config->getVariable("physics", "combat_speed_boost", "100"));
    //same capped big speed as combat...else different
    return (!combat_mode) ? combat_mode_mult * max_combat_speed : max_combat_ab_speed;
}
```

But from vegastrike.config:
`<var name="combat_speed_boost" value="1000"/>`

royfalk (approved):
Approved.

stephengtuggy:
> > I'm positive that the old code was x1000. I remember it distinctly.
> 
> You were right. From computer.cpp in 0.9.x:
> 
> ```
> float Computer::max_speed() const {
>     static float
>             combat_mode_mult = XMLSupport::parse_float(vs_config->getVariable("physics", "combat_speed_boost", "100"));
>     return (!combat_mode) ? combat_mode_mult * max_combat_speed : max_combat_speed;
> }
> 
> float Computer::max_ab_speed() const {
>     static float
>             combat_mode_mult = XMLSupport::parse_float(vs_config->getVariable("physics", "combat_speed_boost", "100"));
>     //same capped big speed as combat...else different
>     return (!combat_mode) ? combat_mode_mult * max_combat_speed : max_combat_ab_speed;
> }
> ```
> 
> But from vegastrike.config: `<var name="combat_speed_boost" value="1000"/>`

Aha! Good to know where that value was coming from.

stephengtuggy:
The travel mode multiplier fix isn't quite ready. The SPEC Capacitor II fix is ready, on the other hand. I'm splitting the latter out into its own PR, #167 .

config.json
```diff
@@ -2,44 +2,44 @@
 "general": {},
 
 "graphics": {
-    "resolution_x": 2560,
-    "resolution_y": 1600,
+    "resolution_x": 1920,
+    "resolution_y": 1080,
```
BenjamenMeyer, line 6: yes we should have a smaller screen by default; even 1080x800 might be better for a default

BenjamenMeyer (approved):

config.json
stephengtuggy: How about 1280x960? That should work almost everywhere, while giving us a little more screen real estate than we've had in the past by default.

BenjamenMeyer: that should be fine. Just so long as its something that fits well in a window.

stephengtuggy: @royfalk what are your thoughts on this?

evertvorster: I'm not Roy, of course, but I do have some thoughts. 
Firstly, 1920x1080 seems to be a sane default. 
It's not related to the travel mode multiplier though, so maybe have a new pr for this?

Looking down the road we would like to have this configure-able somehow, and surviving game updates, so there might be some things to discuss in the PR.

royfalk: Agree with Evert.
Full HD has been a standard for a decade now.
Also on the bit about a new  PR and persistence of the settings.

stephengtuggy: @evertvorster @royfalk Good suggestions. I have split the resolution change off into its own PR: #170

stephengtuggy: And the new Issue is: https://github.com/vegastrike/Assets-Masters/issues/68

BenjamenMeyer: For reference, my laptop display is 1920x1080 to start with.
We should have a sane default for folks with a single monitor that can fit in a window on that monitor.

stephengtuggy: > For reference, my laptop display is 1920x1080 to start with. We should have a sane default for folks with a single monitor that can fit in a window on that monitor.

Sounds reasonable. My laptop display is that same resolution, FWIW.

stephengtuggy:
https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1086 actually fixes the issue in practice. This PR just corrects the value in the new-style config, which isn't used yet.

BenjamenMeyer (approved):

config.json
BenjamenMeyer: Thanks - nice that #170 split out just the resolution change

PR #170, opened 2025-04-01
config.json: Change default resolution

stephengtuggy:
Code Changes:

Issues:
- Fixes https://github.com/vegastrike/Assets-Masters/issues/68

Purpose:
- What is this pull request trying to do? Change the config.json default resolution to a smaller value for broader compatibility.

### DIFF
--- a/config.json
+++ b/config.json
@@ -2,8 +2,8 @@
 "general": {},
 
 "graphics": {
-    "resolution_x": 1920,
-    "resolution_y": 1080,
+    "resolution_x": 1280,
+    "resolution_y": 960,
     "screen": 0
 },
   
--- a/config.json
+++ b/config.json
@@ -2,8 +2,8 @@
 "general": {},
 
 "graphics": {
-    "resolution_x": 2560,
-    "resolution_y": 1600,
+    "resolution_x": 1920,
+    "resolution_y": 1080,
     "screen": 0
 },
   


stephengtuggy:
It sounds like several folks want the default resolution to be 1920x1080. I agree that this is a very common screen size. However, when running Vega Strike inside a VM (which I have often done for testing purposes), I find that I need to use a smaller resolution in order to fit inside the VM window. The vertical resolution is the main limiting factor, most of the time. 960 pixels should work in almost all cases. Hence my vote for a default resolution of 1280x960 -- although we could make it wider if we want.

stephengtuggy:
And yes, definitely we will want to preserve the resolution between program updates. I think there should be a copy of config.json in the `$HOME/.vegastrike` directory that takes precedence over the copy of the file in the Assets directory. Much the same way that the vegastrike.config file works today.

evertvorster (approved):
Looks good to me!

evertvorster (approved):
This is a small and simple change, and therefore approved.

stephengtuggy:
> This is a small and simple change, and therefore approved.

@evertvorster Cool. I just want to make sure we've completed discussing this adequately before merging the change.

BenjamenMeyer (approved):

evertvorster (approved):
All good here

PR #174, opened 2025-04-18
Fix some settings values

stephengtuggy:
Code Changes:
- [X] (Largely) Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues Fixed:
- ...

Known Issues Remaining:
- ...

Purpose:
- What is this pull request trying to do? Fix/update several configuration() setting values, to go with the latest Game Engine PRs.

### DIFF
--- a/config.json
+++ b/config.json
@@ -352,8 +352,8 @@
 
         "shield": {
             "energy_source": "energy",
-            "maintenance_factor": 0.1,
-            "regeneration_factor": 0.1
+            "maintenance_factor": 0.05,
+            "regeneration_factor": 0.05
         }
 
     },
@@ -1296,13 +1296,13 @@
         "player_warp_energy_multiplier": 0.12,
         "use_warp_energy_for_cloak": true,
         "warp_energy_multiplier": 0.12,
-        "warp_memory_effect": 0.9,
+        "warp_memory_effect": 0.5,
         "warp_multiplier_max": 300000000.0,
         "warp_multiplier_min": 1.5,
         "warp_ramp_down_time": 0.5,
         "warp_ramp_up_time": 5.0,
         "warp_stretch_cutoff": 500000.0,
-        "warp_stretch_decel_cutoff": 500000.0,
+        "warp_stretch_decel_cutoff": 2000000.0,
         "fov_link": {
             "reference_kps": 0.0,
             "smoothing": 0.4,
--- a/config.json
+++ b/config.json
@@ -999,7 +999,7 @@
         "auto_landing_warning_distance": 350.0,
         "auto_pilot_compensate_for_interdiction": false,
         "auto_pilot_no_enemies_distance_multiplier": 4.0,
-        "auto_pilot_planet_radius_percent": 0.2,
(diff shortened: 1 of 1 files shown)


vincele:
I don't know what this is fixing, nothing is explained. Why are those values better ?

config.json
```diff
@@ -352,8 +352,8 @@
 
         "shield": {
             "energy_source": "energy",
-            "maintenance_factor": 0.1,
-            "regeneration_factor": 0.1
+            "maintenance_factor": 0.05,
+            "regeneration_factor": 0.05
```
stephengtuggy, line 356: My starting Llama’s shields weren’t recharging. They were taking too much power to maintain, let alone replenish. These two settings changes fix that.

config.json
```diff
@@ -999,7 +999,7 @@
         "auto_landing_warning_distance": 350.0,
         "auto_pilot_compensate_for_interdiction": false,
         "auto_pilot_no_enemies_distance_multiplier": 4.0,
-        "auto_pilot_planet_radius_percent": 0.2,
+        "auto_pilot_planet_radius_percent": 0.5,
```
stephengtuggy, line 1002: This change causes the ship to drop out of autopilot at the correct distance from the target planet, just before docking distance. Although perhaps 0.49 would be an even better setting.

config.json
```diff
@@ -1296,13 +1296,13 @@
         "player_warp_energy_multiplier": 0.12,
         "use_warp_energy_for_cloak": true,
         "warp_energy_multiplier": 0.12,
-        "warp_memory_effect": 0.9,
+        "warp_memory_effect": 0.5,
```
stephengtuggy, line 1299: This setting controls how much of your ship’s speed is saved and restored when dropping out of SPEC and then reentering SPEC mode. This setting was much too high; 0.5 seems to be a better value.

config.json
```diff
@@ -1296,13 +1296,13 @@
...
         "warp_multiplier_max": 300000000.0,
         "warp_multiplier_min": 1.5,
         "warp_ramp_down_time": 0.5,
         "warp_ramp_up_time": 5.0,
         "warp_stretch_cutoff": 500000.0,
-        "warp_stretch_decel_cutoff": 500000.0,
+        "warp_stretch_decel_cutoff": 2500000.0,
```
stephengtuggy, line 1305: This change reduces the amount by which the autopilot overshoots the target, with stations, for example. Now, if it overshoots at all, it’s frequently only by 5 km or so, instead of 50,000 or more.

stephengtuggy:
> I don't know what this is fixing, nothing is explained. Why are those values better ?

Good point. I should have documented my changes better. I have now done that.

vincele:
That would have been nice in the commit messages, but OK...

stephengtuggy:
> That would have been nice in the commit messages, but OK with me...

Yes, that would have been better. I'll try to remember that next time

royfalk (approved):

PR #182, opened 2025-05-19
Fix vsUTCS Start Menu Item for 0.9.x

stephengtuggy:
Code Changes:
- [X] This is an installer change only, and can't be tested very well until a new installer is built.

Issues Addressed:
- Fixes https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/1130 (hopefully)

Known Issues Remaining:
- No new ones that I'm aware of

Purpose:
- What is this pull request trying to do? Fix the set of Vega Strike menu item(s) installed by the installers
- What release is this for? 0.9.1
- Is there a project or milestone we should apply this to? 0.9.x

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -213,10 +213,10 @@ IF (CMAKE_SYSTEM_NAME STREQUAL Windows)
     SET(CPACK_NSIS_CONTACT "maintainers@vega-strike.org")
     SET(CPACK_NSIS_MODIFY_PATH OFF)
     SET(CPACK_NSIS_CREATE_ICONS_EXTRA
-        "CreateShortCut '$SMPROGRAMS/$STARTMENU_FOLDER/Vega Strike Upon the Coldest Sea.lnk' 'C:/Program Files/VegaStrike-0.9/bin/vegastrike-engine.exe' -D\"$INSTDIR\\share\\vegastrike\""
+        "CreateShortCut '$SMPROGRAMS\\\\$STARTMENU_FOLDER\\\\Vega Strike Upon the Coldest Sea.lnk' 'C:\\\\Program Files\\\\VegaStrike-0.9\\\\bin\\\\vegastrike-engine.exe' -D\"$INSTDIR\\share\\vegastrike\""
     )
     SET(CPACK_NSIS_DELETE_ICONS_EXTRA
-        "Delete '$SMPROGRAMS/$START_MENU/Vega Strike Upon the Coldest Sea.lnk'"
+        "Delete '$SMPROGRAMS\\\\$START_MENU\\\\Vega Strike Upon the Coldest Sea.lnk'"
     )
     SET(CPACK_GENERATOR "NSIS64")
     SET(CPACK_PACKAGE_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/packages")


CMakeLists.txt
```diff
@@ -211,8 +211,13 @@ IF (CMAKE_SYSTEM_NAME STREQUAL Windows)
     SET(CPACK_NSIS_HELP_LINK "https://vega-strike.org")
     SET(CPACK_NSIS_URL_INFO_ABOUT "https://vega-strike.org/about")
     SET(CPACK_NSIS_CONTACT "maintainers@vega-strike.org")
-    SET(CPACK_NSIS_MODIFY_PATH ON)
-
+    SET(CPACK_NSIS_MODIFY_PATH OFF)
+    SET(CPACK_NSIS_CREATE_ICONS_EXTRA
+        "CreateShortCut '$SMPROGRAMS\\\\$STARTMENU_FOLDER\\\\Vega Strike Upon the Coldest Sea.lnk' 'C:\\\\Program Files\\\\VegaStrike-0.9\\\\bin\\\\vegastrike-engine.exe' -D\"$INSTDIR\\share\\vegastrike\""
```
BenjamenMeyer, line 216: could we use the variables from L26-L28 to build the path here?

Also, we do we need to use `\`? Windows will typically use `/` just fine too, and would make it cleaner:

```
"CreateShortCut '$SMPROGRAMS/$STARTMENU_FOLDER/Vega Strike Upon the Coldest Sea.lnk' 'C:/Program Files/VegaStrike-0.9/bin/vegastrike-engine.exe' -D\"$INSTDIR/share/vegastrike\""
```

stephengtuggy: Yeah, the forward slashes didn't work. Backslashes it is.

royfalk: Supposedly it works if you use it within a powershell script. However, I'm not convinced it's worth the effort. Are we even requiring powershell on windows for Vegastrike?

```
$shell = New-Object -comObject WScript.Shell
$shortcut = $shell.CreateShortcut($shortcutPath)
$shortcut.TargetPath = $targetPath
$shortcut.Arguments = ""  # You can add arguments here if needed
$shortcut.Save()
```

royfalk (approved):
So, you're basically doing three things here:
1. Remove the option to change install location.
2. Create a shortcut manually with the -d parameter.
3. Delete the shortcut without the -d.

Seems straightforward.

CMakeLists.txt
stephengtuggy: We are not requiring PowerShell on Windows for Vega Strike. That's why I didn't want to use anything PowerShell-specific.

stephengtuggy:
> So, you're basically doing three things here:
> 
> 1. Remove the option to change install location.
> 2. Create a shortcut manually with the -d parameter.
> 3. Delete the shortcut without the -d.
> 
> Seems straightforward.

Almost.

1. I am removing the option to change the install location. As a side effect, I am removing the option to create a desktop shortcut, since if that shortcut was created, it wouldn't be correct.
2. Yes, create a custom shortcut on the Start Menu with the -d parameter on install.
3. Delete the custom shortcut on uninstall.

stephengtuggy:
This NSIS stuff is pretty complicated. And goofy, seems to me.

stephengtuggy:
Well, unfortunately, CPack's NSIS config does not allow customizing the installation folder without the option to create a desktop shortcut. And the desktop shortcut is not customizable, so it would be a broken shortcut.

If you know of a way around this, by all means, let me know.

CMakeLists.txt
BenjamenMeyer: @royfalk I've done in command.com and cmd.exe for years. But not going to force it.

@stephengtuggy thanks for tryiing

PR #184, opened 2025-06-05
Fix issue 74

royfalk:
Fixes https://github.com/vegastrike/Assets-Masters/issues/74

- Count facets when value is 0
- Delete 0 value armor/shield

Code Changes:
- [x] Have the PR Validation Tests been run? minimal. 
    - Changes made to ship assets. Visually checked part of ships.json to ensure correctness.
    - Verified issue is fixed for Gawain.blank
    - Short game play test fighting pirates

### DIFF
--- a/units/parser.py
+++ b/units/parser.py
@@ -119,7 +119,7 @@ with open("units_description.csv") as description_file:
      
      
 json_object = json.dumps(units, indent = 4)
-with open("units.json", "w") as json_file:
+with open("units_old.json", "w") as json_file:
     json_file.write(json_object)
 #print(json_object)
               
--- a/units/ships.json
+++ b/units/ships.json
@@ -46,7 +46,8 @@
                 "Light": "{supernova.bmp.bfxm;-0.249996;0.00000e+0;-1.83758;1.5;;;;;}{supernova.bmp.bfxm;0.250004;0.00000e+0;-1.83758;1.5;;;;;}",
                 "Mounts": "{;;;LIGHT SPECIAL;-4.07186e-6;-0.150000;1.99709;;;-1.06144e-5;0.00000e+0;2.00000;0.00000e+0;1.00000;0.00000e+0;1;1}",
                 "Upgrade_Storage_Volume": "16",
-                "armor": "6.0"
+                "armor": "6.0",
+                "shield_facets": "2"
             },
             "units": [
                 {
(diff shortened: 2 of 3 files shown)


units/parser.py
```diff
@@ -119,7 +119,7 @@ def parseLine(text):
      
      
 json_object = json.dumps(units, indent = 4)
-with open("units.json", "w") as json_file:
+with open("units_old.json", "w") as json_file:
```
stephengtuggy, line 122: ? Why are we writing to units_old.json, rather than to units..json?

units/ships.json
```diff
@@ -560,8 +560,7 @@
                         "Prohibited_Upgrades": "mult_shady_moreupgrade;mult_shady_morethrust;mult_shady_moreshields;mult_shady_moregunrecharge;mult_shady_moreshieldrecharge;mult_shady_moreturning;upgrades/Overdrive;upgrades/Capacitors:1;upgrades/Reactors:1;upgrades/Armor:1;upgrades/Sensors:1;upgrades/Shield_Systems:1;upgrades/Passenger_Quarters:1;upgrades/Repair_Systems:1;upgrades/ECM_Systems:1",
                         "Mounts": "{KtekBeam;;;Light Medium Heavy autotracking;-9;-.25;-.75;;;;;;;;;1;1}{KtekBeam;;;Light Medium Heavy autotracking;9;-.25;-.75;;;;;;;;;1;1}{Ktek;;;Light;-4.5;-.6;12.25;;;;;;;;;1;1}{Ktek;;;Light;4.5;-.6;12.25;;;;;;;;;1;1}{MiniGravThumper;8;200;special-missile;0;-0.5;0;;;;;;;;;1;1}",
                         "armor": "1400.0",
-                        "shield": "412.5",
-                        "shield_facets": "4"
```
stephengtuggy, line 564: Just to make sure I understand correctly: We're moving `"shield_facets"` from the `"units"` section to the `"data"` section? And the same with the other ships?

stephengtuggy:
Hopefully someone can run most, if not all, of the standard play tests as well. Other than that and the items I noted, this is looking good.

units/ships.json
royfalk: The file is hierarchical.
Attributes are always in the data section.
Ships inheriting from this ship are in the units array with their own data section (where different).

Because shield_facets was only generated for ships with shield, it only appeared for "*" or "*.milspec" for example.
Once we fixed this, the value is in the parent "*.blank" and therefore removed from the inheriting units.

stephengtuggy: I see. Sounds good.

units/parser.py
stephengtuggy: Then maybe call it something like `units_temp.json` or `units_diff.json`?

stephengtuggy (approved):
LGTM

PR #198, opened 2025-07-02
Support for engine PR 1233

royfalk:
- Fix mult_shady_moreupgrade__upgrades. Note reduction in mass and volume as this is really a service. However, did not want to provide 0 cargo, as that may break the game.
- Add enslave/free key bindings

This PR provides the necessary assets changes to support https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1233.

Code Changes:

Issues:
- None identified

### DIFF
--- a/config.json
+++ b/config.json
@@ -128,7 +128,9 @@
         "ConsoleKeys::BringConsole": {"key": "function-1", "modifier": "ctrl"},
         "SetVelocityRefKey": {"key": "keypad-7", "modifier": "none"},
         "SetVelocityNullKey": {"key": "keypad-1", "modifier": "none"},
-        "Cockpit::Quit": {"key": "esc", "modifier": "none"}
+        "Cockpit::Quit": {"key": "esc", "modifier": "none"},
+        "FreeSlaves": {"key": "f", "modifier": "ctrl"},
+        "Enslave": {"key": "e", "modifier": "ctrl"}
     },
 
     "mouse": {
--- a/master_component_list.json
+++ b/master_component_list.json
@@ -2044,8 +2044,8 @@
         "file": "mult_shady_moreupgrade",
         "categoryname": "upgrades/Shady_Mechanic",
         "price": "32000",
-        "mass": "10",
-        "volume": "50",
+        "mass": "1",
+        "volume": "0.1",
         "description": "@cargo/cargo_hud.image@Trades cargo space for upgrade room.",
         "upgrade": true
     },
--- a/units/units.json
+++ b/units/units.json
@@ -11489,7 +11489,8 @@
         "Name": "More Upgrade Volume",
         "Object_Type": "Upgrade_Improvement",
         "Textual_Description": "\"@cargo/cargo_hud.png@Trades cargo space for upgrade room\"\n",
-        "Hold_Volume": "-50"
+        "Hold_Volume": "-20",
+        "Upgrade_Storage_Volume": "12"
     },
     {
         "Key": "hud_repair__upgrades",
--- a/vegastrike.config
+++ b/vegastrike.config
@@ -284,6 +284,8 @@ Farther down, you will see "#GeomLow". Currently, it's line 213. You'll find mor
 
 
     <bind key="esc" modifier="none" command="Cockpit::Quit"/>
+    <bind key="f" modifier="ctrl" command="FreeSlaves"/>
+    <bind key="e" modifier="ctrl" command="Enslave"/>
 
 <!-- #no_mouse -->
     <bind mouse="0" player="1" button="0" modifier="none" command="FireKey" />


master_component_list.json
```diff
@@ -2044,8 +2044,8 @@
         "file": "mult_shady_moreupgrade",
         "categoryname": "upgrades/Shady_Mechanic",
         "price": "32000",
-        "mass": "10",
-        "volume": "50",
+        "mass": "1",
+        "volume": "0.1",
```
stephengtuggy, line 2048: I don't like this change. If I understand correctly, it will reduce ethe amount of extra upgrade room you can obtain to almost nothing. I don't think we should be making these kinds of changes to the functionality of the game until we fix the critically-broken things like memory-management and threading issues.

config.json
```diff
@@ -128,7 +128,9 @@
         "ConsoleKeys::BringConsole": {"key": "function-1", "modifier": "ctrl"},
         "SetVelocityRefKey": {"key": "keypad-7", "modifier": "none"},
         "SetVelocityNullKey": {"key": "keypad-1", "modifier": "none"},
-        "Cockpit::Quit": {"key": "esc", "modifier": "none"}
+        "Cockpit::Quit": {"key": "esc", "modifier": "none"},
+        "FreeSlaves": {"key": "f", "modifier": "ctrl"},
+        "Enslave": {"key": "e", "modifier": "ctrl"}
```
stephengtuggy, line 133: This part should be OK.

master_component_list.json
royfalk: There seems to be a misunderstanding here. The game does not have the concept of a service. It therefore treats anything you can buy as a cargo item. Add more upgrade space or add more cargo space are not strictly upgrades. These are services done to the ship to modify it.

This creates all sorts of interesting problems. If you happen to somehow buy or acquire a "more upgrade volume", how much space would it take in your cargo hold?

I solved this by making the mass 1 and the volume 0.1. 

In any case, when you upgrade or downgrade, you buy the cargo upgrade and then the attributes in units.json are applies to the ship. Actual behavior of the game is unchanged (or nearly so).

stephengtuggy: Hmm. I'll have to think about this some more.

stephengtuggy: Nope, I am not OK with this change. You are neutering important game functionality here. Please refrain.

stephengtuggy (changes requested):
See comments above.

master_component_list.json
royfalk: I'm trying to understand what it is you are objecting to. 

> Nope, I am not OK with this change. You are neutering important game functionality here. Please refrain.

Can you please specify what functionality is now neutered?
Simply put, if you buy a `more upgrade`, it will reduce cargo hold size and increase upgrade space. 
If you buy 'more cargo' it will increase cargo hold size and reduce upgrade space.

> If I understand correctly, it will reduce the amount of extra upgrade room you can obtain to almost nothing. 

No. More upgrade will increase upgrade space by 12 and reduce cargo hold by 20. I'm not sure if this was the original number but it seems like a reasonable ratio.
Add cargo (not changed) will increase hold by 12 and reduce upgrade by 20.

If you feel the ratio of 6 to 10 is unreasonable, I'm open to any other.

evertvorster: Hmm, why would the total space decrease for upgrade space improvements? Conceptually, you are just moving a dividing wall between your upgrade space and cargo space, so each cubic feet of space you lose on the one side of the wall should be added to the other side of the wall. 

Now, cargo space improvements may not even necessarily remove any space from upgrade space, as you either 1. utilize previously unused void spaces in your ship as cargo space, or 2. weld some container on the outside, like a roof carrier on a car.

royfalk: I'm currently wearing my developer hat. As such, I'm putting a reasonable value in terms of gameplay.

This is meant to support an engine PR by providing A value for it to use. 

Actual question of whether it should be 1:1 or 6:10 can be decided by consensus or by someone submitted another PR.

Same thing if you want to add a container on the outside. However, that adds additional questions: 
1. Adds to mass and not upgrade space - probably.
2. Outside armor, how to model?
3 Restrict only to specific large cargo ships?!

evertvorster: For upgrade volume, IIRC, there was a description that stated "using previously unused void spaces for upgrades", which means that when buying this type of upgrade volume, the cargo volume should remain unchanged. This would only add a small amount of upgrade volume. Less on smaller ships, more on larger ones. It can only be applied once. 

In theory you could also trade some cargo space for upgrade space by simply installing your upgrades in the cargo area, or moving a wall... in this case the total volume of cargo + upgrade volume should stay constant.

For cargo volume upgrades, we are typically talking about much bigger increase in total volume, which would mean physically altering the ship to have a bigger cargo volume. 
On oceangoing vessels that is achieved by cutting the ship in half and adding in a section of ship that sort of blends in with the rest in the middle, and welding the ends back on to this new section. The modelling implications were hand-waved away, but yes, this should add mass to the ship. And be very, very expensive. Almost the price of a new ship of the same type.

stephengtuggy: OK, I guess I misunderstood. The `1` and `0.1` values do not reflect how much upgrade or hold volume these upgrades add or subtract.

If I found this difficult to understand, I suspect that other people will, too. We'd better document this carefully, if we do make this change.

I still don't understand why we need to make this change though.

`The game does not have the concept of a service` -- Excuse me? What about `Basic Repair and Refuel`?

units/units.json
```diff
@@ -11489,7 +11489,8 @@
         "Name": "More Upgrade Volume",
         "Object_Type": "Upgrade_Improvement",
         "Textual_Description": "\"@cargo/cargo_hud.png@Trades cargo space for upgrade room\"\n",
-        "Hold_Volume": "-50"
+        "Hold_Volume": "-20",
+        "Upgrade_Storage_Volume": "12"
```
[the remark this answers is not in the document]
stephengtuggy: I think a 1:1 ratio would make better sense, personally.

master_component_list.json
royfalk: > I still don't understand why we need to make this change though.

We've been moving away from the games old paradigm of treating a ship/unit as a struct of attributes and upgrading/downgrading by comparing it to another unit instance.

It's really elegant in theory but gets complicated and ugly in a hurry.
To add support for multiple instances, upgrades need to have a `mult_ prefix`.
Also, good luck unit testing this mess.

So, we've moved to components. However, the upgrade space isn't a component. It's what you put the components in. Which makes it a bit more complicated to implement cleanly.
Therefore I wrote this PR which is a stop gap measure and just implements an upgrade/downgrade of cargo/upgrade.


>> The game does not have the concept of a service 
> -- Excuse me? What about Basic Repair and Refuel?

That's a hard coded piece of cruft stuck right in the middle of `basecomputer.cpp`. I would consider a service to be something you can purchase at a service center but cannot sell and also cannot find floating around in space.
As all upgrades are currently cargo, it's theoretically possible to destroy a ship and see its `shield01` floating.
Theoretically, as the current code converts all upgrades to `space salvage`.

royfalk: Also

> OK, I guess I misunderstood. The 1 and 0.1 values do not reflect how much upgrade or hold volume these upgrades add or subtract.

Correct. That's because the game stores attributes in two separate places - cargo, for when you buy something and master_parts, for when you install the thing. Of course, you also buy the corresponding cargo and place it in your "cargo hold" but marked as upgrade. 

> If I found this difficult to understand, I suspect that other people will, too. We'd better document this carefully, if we do make this change.

I'll add a comment to the JSON.

royfalk: I'll start work on something. Created https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/1251 to track.

BenjamenMeyer: @royfalk thanks!

royfalk:
@stephengtuggy as it says in the summary, this PR provides the necessary changes to the assets in order to support #1233.

As you've merged that, we either merge this as well or unmerge 1233.

stephengtuggy:
> @stephengtuggy as it says in the summary, this PR provides the necessary changes to the assets in order to support #1233.
> 
> As you've merged that, we either merge this as well or unmerge 1233.

@royfalk OK. Sorry I missed that part. We can merge this, then fine-tune it later, I suppose.

PR #213, opened 2025-09-16
Fix Oswald attacking for no reason (1266)

royfalk:
Fix Oswald respawning after dying (1310).

**Bug triage**
- 1266 was due to Oswald using SPEC when getting nearer. Probably due to our tinkering with the auto-pilot/configuration. The script checked the shields weren't down to decide if Oswald is attacked, which caused the bug.
- 1310 was due to the devs not considering the players can kill Oswald. He flew a powerful ship and you really can't defeat him with Llama.begin. Killing him did not remove the quest, which caused the bug.

Fixes https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/1266
Fixes https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/1310

**Code Changes:**
- [x] Have the PR Validation Tests been run? 

Tested:
- Able to complete tutorial without Oswald losing shields and getting angry.
- Killing Oswald removes quest.

Issues:
- Multiple other bugs in the tutorial quest, all minor.

### DIFF
--- a/modules/quests/quest_tutorial.py
+++ b/modules/quests/quest_tutorial.py
@@ -51,6 +51,8 @@ COMPLETE_TUTORIAL4 = 8
 COMPLETE_TUTORIAL5 = 10
 STAGE_DECLINE = 98
 
+oswald_launched = False
+
 # the class that will be executed
 class quest_tutorial (quest.quest):
     # initialize quest variables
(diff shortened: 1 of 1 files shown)


BenjamenMeyer:
> He flew a powerful ship and you really can't defeat him with Llama.begin.

FYI - I often use Dumbfire Missiles and kill him with the initial Llama ship (in the initial supply of the ship at the start of the game). :)

royfalk:
> FYI - I often use Dumbfire Missiles and kill him with the initial Llama ship

Fair enough. I'm not sure you could have done that 20 years ago. Either that, or they just didn't think about it and so the bug...

> seems fair - do we have anyone else that has play tested this yet?

Give them a few minutes. It hasn't been up for an hour yet.

evertvorster (changes requested):
This gets rid of Oswald respawning.
Oswald also does not attack when starting a new campaign. 
However, he also does not start the training.... 

Could you please test again to see that he actually starts the training when starting a new campaign? On my last test of this, he just sat there, motionless and did not actually approach me to start the tutorial.

royfalk:
> However, he also does not start the training....

You should have approached him :)
In any case, it's fixed now.

PR #240, opened 2026-05-07
game_config.py: Apply some PyCharm suggestions

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- N/A

Purpose:
- What is this pull request trying to do? Apply some PyCharm IDE suggestions to game_config.py
- Is there a project or milestone we should apply this to? 0.10.x, VS Configuration Tooling

### DIFF
--- a/python/settings_app/game_config.py
+++ b/python/settings_app/game_config.py
@@ -1,10 +1,8 @@
 # A class to manage the configuration, both from user folder and from the assets folder.
 
+import copy
 import json
 import os
-import getpass
-import platform
-import copy
 
 import app_config as ac
 
@@ -25,10 +23,10 @@ CONFIG_FILE_NAME = "config.json"
 
 
 
-class ConfigBranch():
-    def __init__(self, parent, key: str, asset_dict: dict, user_dict: dict = None):
+class ConfigBranch:
+    def __init__(self, parent, key: str | None, asset_dict: dict, user_dict: dict = None):
         self.parent: ConfigBranch = parent
-        self.key: str = key
+        self.key: str | None = key
         self.value = {}
         self.dirty = []  # A list of all dirty leaves
 
@@ -36,7 +34,7 @@ class ConfigBranch():
             # Branch
             if isinstance(value, dict):
                 # We have a user branch
-                if user_dict != None and key in user_dict:
+                if user_dict is not None and key in user_dict:
                     self.value[key] = ConfigBranch(parent = self, key = key, asset_dict = value, user_dict = user_dict[key])
                 # No user branch
                 else:
@@ -44,7 +42,7 @@ class ConfigBranch():
             # Leaf
             else:
                 # We have a user leaf
-                if user_dict != None and key in user_dict:
+                if user_dict is not None and key in user_dict:
                     self.value[key] = ConfigLeaf(parent = self, key = key, value = user_dict[key], original_value = value)
                 # No user branch
                 else:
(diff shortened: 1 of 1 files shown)


python/settings_app/game_config.py
```diff
@@ -25,26 +23,26 @@
 
 
 
-class ConfigBranch():
```
stephengtuggy, line 28: PyCharm indicated that these empty parentheses were unnecessary.

python/settings_app/game_config.py
```diff
@@ -25,26 +23,26 @@
...
+        self.key: str | None = key
         self.value = {}
         self.dirty = []  # A list of all dirty leaves
 
         for key, value in asset_dict.items():
             # Branch
             if isinstance(value, dict):
                 # We have a user branch
-                if user_dict != None and key in user_dict:
+                if user_dict is not None and key in user_dict:
```
stephengtuggy, line 37: Apparently `is None` or `is not None` is a better way to compare with `None`.

python/settings_app/game_config.py
```diff
@@ -92,16 +90,16 @@ def get_object(self, key: list[str]):
             return None
         
     def get_changes_dictionary(self):
-        dict = {}
+        return_value = {}
```
stephengtuggy, line 93: PyCharm complained that `dict` conflicted with the type name, so I renamed it.

python/settings_app/game_config.py
```diff
@@ -1,10 +1,8 @@
 # A class to manage the configuration, both from user folder and from the assets folder.
 
+import copy
 import json
 import os
-import getpass
-import platform
```
stephengtuggy, line 6: I guess these imports aren't used.

python/settings_app/game_config.py
BenjamenMeyer: correct - `dict` is a type itself

python/settings_app/game_config.py
BenjamenMeyer: yes. The reason has to do with what `None` is in Python, or rather CPython - the primary interpreter for python (technically it was supposed to be a reference implementation but it became the default implementation). Any how...

In Python everything is an `Object` type derived from `PyObject` (C API) or `object` (Python) and everything can technically be created at run time. However for optimizations there are a number of objects (numbers, 0 to 255, the basic alphabet, true, false, etc) that are instantiated as object singletons. The `None` object is one of those singleton objects.

Now when you do a comparison with `=` and `!=` it performs a _value_ comparison - does the object on the left and the object on the right contain the same value? This is supported via the `__eq__()` method on an object.

However, when you do an `is` or `is not` comparison it perform what is essentially a _pointer_ comparison - is the object on the left the _same_ as the object on the right, which is only true if it is the same exact object (same address, etc). If it is the same object then it also has the same value. Thus it is O(1) runtime.

And of course comparing a pointer value (address) is far faster than retrieving the pointer, decoding it, and then calling its comparison method (`left.__eq__(right)`) which then does some non-constant runtime call.

python/settings_app/game_config.py
BenjamenMeyer: yes, you should either remove them or use `ConfigBranch(object)`.

In Py2 there was multiple ways to construct an object and not all of them derived nicely from `object`, and using `ObjectName(object)` was necessary for Py3 compatibility.

In Py3 (which is what all our code now runs under) everything cleanly derives from `object` which makes `ObjectName(object)` unnecessary and you can just do `ObjectName` instead.

BenjamenMeyer (approved):

PR #247, opened 2026-07-12
Hi-res cargo images

SGD1953:

Please answer the following:

Code Changes:
- [X] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues:
First part of https://github.com/vegastrike/Assets-Masters/issues/103

Purpose:
This standardised all cargo images to 512x512 so they are less blurry on higher resolutions


BenjamenMeyer (commented):
We'll have to pull this down to check.

@royfalk @SGD1953 @stephengtuggy we might want to rethink using the `.image` extension for all the PNGs since it keeps us from being able to view it easiy on GitHub so reviews are more work. I understand the decision when it was made (before our time); but tools have changed since then so it may be wiser to just use the PNG extension instead. We're not going to have many file types any way; but for any images (PNG, MNG) and videos (MP4, MOV, and WebM; H264 or AAC encoding) it may be wise to stick with extensions that Github recognizes to make reviews easier.

BenjamenMeyer (approved):
pulled down and looked at this locally - looks good.

SGD1953:
In principle, there is a sprite folder and a texture folder, so the purpose of the file is clear (funny enough, the cargo sprites are under textures). Therefore I agree that we should just use the .dds extension. The new scripts would make this fairly easy for the most part, bfxm files will have to be updated, but that should not be an issue as all textures are in the first line.

PR #248, opened 2026-07-12
Fix 0.10.0 alpha installers

### LINKED ISSUE #105: Windows 2025 build failing because NSIS is missing
The GitHub Actions Windows 2025 images [do not include NSIS](https://github.com/actions/runner-images/issues/11754). We will need to install it ourselves as part of the installer workflow, I guess.

stephengtuggy:

Please answer the following:

Code Changes:
- [x] This is a CI/build system change only

Issues:
- Fixes https://github.com/vegastrike/Assets-Masters/issues/105

Purpose:
- What is this pull request trying to do? On Windows 2025, install NSIS before trying to create the installer, as NSIS is no longer installed automatically on the GitHub runner image. Make other changes as needed to fix the installer builds.
- What release is this for? 0.10.x and following

### DIFF
--- a/.github/workflows/Windows-Release.yml
+++ b/.github/workflows/Windows-Release.yml
@@ -26,8 +26,10 @@ jobs:
         include:
           - os:              windows-2022
             is-release:      1
+            install-nsis:    0
           - os:              windows-2025
             is-release:      1
+            install-nsis:    1
 
     env:
       TAG_NAME: '${{ github.ref_name }}'
@@ -52,6 +54,12 @@ jobs:
 
       - name: Install CMake
         uses: lukka/get-cmake@f5b8fbb4d77cec1acc5a5f9f0df4beffaf5d98d9 #v4.3.4
+
+      - name: Install NSIS
+        if: ${{ matrix.install-nsis }}
+        uses: repolevedavaj/install-nsis@1ab2f7ba2fe9408db612029be0bb3c0088b0d3e9 # v1.2.1
+        with:
+          nsis-version: '3.12'
 
       - name: Build it
         run: script/build.ps1
--- a/.github/workflows/Windows-Release.yml
+++ b/.github/workflows/Windows-Release.yml
@@ -61,10 +61,10 @@ jobs:
         with:
           nsis-version: '3.12'
 
-      - name: Build it
+      - name: Build the code
         run: script/build.ps1
 
-      - name: Package it
+      - name: Package the code
         run: script/package.ps1
 
       - name: Upload the artifacts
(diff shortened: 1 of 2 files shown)


SGD1953 (approved):

.github/workflows/Windows-Release.yml
```diff
@@ -26,8 +26,10 @@ jobs:
         include:
           - os:              windows-2022
             is-release:      1
+            install-nsis:    0
```
BenjamenMeyer, line 29: is there a reason we don't need it on Windows 2022?

script/package
```diff
@@ -95,18 +99,18 @@ function buildPackages()
 
     while [ $hdiutil_success -ne 1 ] && [ $try_count -lt 8 ]; do
         # Create temporary rw image
-        if make package
+        if cpack -V -G DragNDrop
```
BenjamenMeyer: this right? [DragNDrop](https://cmake.org/cmake/help/latest/cpack_gen/dmg.html) IIRC is for Mac only to create macOS's DMG Image

where `make package` ran the cmake to build it for each platform.

BenjamenMeyer (changes requested):
The `cpack` command is wrong. This won't build all the different package types - just the DMG file for macOS.

A comment of about why we need NSIS installed on Win2025 and not on Win2022 would be good.

.github/workflows/Windows-Release.yml
stephengtuggy: The NSIS software for making installers comes preinstalled on the Windows Server 2022 GitHub Actions runners. Microsoft, in their infinite wisdom, decided to leave it out of the Windows 2025 runner images. Hence why we have to install it ourselves.

script/package
stephengtuggy: Oh, you're right. I believe CPack is the better command to run. But I was forgetting that this script is not mac-only.

I think I saw an error somewhere, in one of the macOS builds yesterday or today, to the effect that CPack didn't know what generator to use. But I think I got mixed up about where that was.

How about if I change this to just `cpack -V`?

stephengtuggy: There. I went ahead and just reverted it to `make package`. It was working fine as it was. Why change it?

BenjamenMeyer: `make package` should build it for each platform based on the CMake configuration when `cmake .` was run... 

`make package` :arrow_right: `make package_source` :arrow_right: `cmake in cpack mode`

Unfortunately I'm not sure how to make it go verbose that way (`cpack -V`) which will certainly give us a lot better logs.

Using `make package` has the benefit of ensuring that the build system is up-to-date with all dependencies resolved as it automatically runs `make all`. Otherwise yes, `cpack -V` would be right.

BenjamenMeyer (approved):

script/package
stephengtuggy: I think you can set the `VERBOSE` environment variable immediately beforehand. I.e.:

```bash
VERBOSE=1 make package
```

PR #256, opened 2026-07-21
Better readability and formatting for player and ship info

SGD1953:
Code Changes:
- [X] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

This adds nicer formatting for ship and player info with aligned numbers.

### DIFF
--- a/python/base_computer/player_info.py
+++ b/python/base_computer/player_info.py
@@ -38,10 +38,10 @@ def get_player_info(names, relations, kills):
         color = get_color(relation)
         
         if faction_kills == 0 or name == 'privateer':
-            text += f"{light_grey}{name}:{end_color} {color}{relation}{end_color}{newline}"
+            text += f"{light_grey}{name:<24}{end_color} {color}{relation:>3}{end_color}{newline}"
         else: 
-            text += f"{light_grey}{name}:{end_color} {color}{relation}, kills: {faction_kills}{end_color}{newline}"
+            text += f"{light_grey}{name:<24}{end_color} {color}{relation:>3}, kills: {faction_kills:>3}{end_color}{newline}"
     
-    text += f"{newline}{bold}Total Kills: {total_kills}{end_bold}{newline}"
+    text += f"{newline}{bold}{"Total Kills":<24} {total_kills:>3}{end_bold}{newline}"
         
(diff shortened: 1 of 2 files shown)


BenjamenMeyer (approved):
Looks reasonable; I haven't run it but it make sense. Good find on the `units_old.json` vs the `ships.json` as well.

SGD1953:
That direct execution does not really work though, the games seems to mix in more properties at runtime which are not in the JSON

BenjamenMeyer:
> That direct execution does not really work though, the games seems to mix in more properties at runtime which are not in the JSON

more context?

FYI - the old configuration tool had a ton of settings in the file, but only exposed a few to the user in the app. Our new tool exposes almost all to the user in the Python settings app; we'll probably figure out a balance later or which should be hidden behind an `advanced config` mode or something. @royfalk can give you a better run down on how the settings get into the new app as there's some auto-generation stuff he developed to support it.

SGD1953:
This PR has nothing to do the the settings app, it is about the "Player Info" and "Ship info" screen in the base computer. My remark was just on the main method in the script which was once meant to allow running it standalone (which is neat if you do a lot of editing and want to instantly validate.

SGD1953:
@BenjamenMeyer can we get this merged?

royfalk:
> That direct execution does not really work though, the games seems to mix in more properties at runtime which are not in the JSON

Yep. That's how I started development of this and then moved it to run from the engine. It's still useful for some rapid development, even if some stuff is broken or incomplete.

SGD1953:
Agreed, might be worthwhile to fix by i.e. just augmenting the data read from file with dummy data so that everything is there.

royfalk:
As for `with open('units/units_old.json', 'r') as file:` the new `ships.json` is hierarchical.
That means that in order to generate a ship, you need to collect values from all json objects above.

I switched to JSON to modernize things and then the file got really big. So I implemented this to reduce the file and this had knock-on effects like not being able to use it here.

royfalk:
Go ahead and merge it.

BenjamenMeyer (approved):

PR #259, opened 2026-07-31
Fix wayland screen detection

evertvorster:

Please answer the following:

Code Changes:

This has very little impact on the game engine, and the code only modifies the screen resolutions that are settable with the app. 

That said, I did play test at 4K resolution. The game plays, but there are some minor layout issues in the HUD and the relatively smaller text in the menus are a little annoying. However, this can be addressed with fixes in the engine to better support multiple resolutions. The issues there have been present for a long time, and are only brought to the front by having a much higher resolution. 

Of course, it also allows you to set any other supported resolution, and you don't HAVE to play at 4K, it just opens the door to setting that resolution through the GUI. 

Issues:
- The settings app mis-detects the proper resolutions available on my Wayland client with a 4K external monitor. 

Purpose:
- - Uses the same SDL dependencies as the config tool to detect screens and resolutions.
- What release is this for?
- - Git master. 


stephengtuggy (commented):
Looks good for the most part. I tested this on my KDE system running in X.Org launched from a terminal, and it worked fine.

I just have two concerns. First, due to the SDL3 dependency, this probably won't work on some of the older Linux distributions that we still support for 0.10.x. It will be fine for 0.11.x though. So I think it's time to split off a 0.10.x branch of Assets-Production before we merge this.

Second, this code uses a very specific filename for the SDL3 library, with a very specific extension. I don't think this will work as is on macOS or Windows.

python/settings_app/sdl_detect.py
```diff
@@ -0,0 +1,56 @@
...
+        ("h", ctypes.c_int),
+        ("pixel_density", ctypes.c_float),
+        ("refresh_rate", ctypes.c_float),
+        ("refresh_rate_numerator", ctypes.c_int),
+        ("refresh_rate_denominator", ctypes.c_int),
+        ("internal", ctypes.c_void_p),
+    ]
+
+def main():
+    sdl = ctypes.CDLL("libSDL3.so.0")
```
BenjamenMeyer: this isn't portable across operating systems; it should work for anything UNIX based; but definitely won't work on Windows.

BenjamenMeyer (dismissed):
yes, the shared library needs to be setup to be OS aware and load the correct one based on OS.

also agree with @stephengtuggy that we should split off 0.10.x before this gets merged.

evertvorster:
Changed the hard-coded libSDL3.so.0 with ctypes.util.find_library like so:

# Before (Linux-only)
sdl = ctypes.CDLL("libSDL3.so.0")

# After (platform-agnostic)
libname = ctypes.util.find_library("SDL3")
if libname is None:
    libname = "libSDL3.so.0"  # fallback for Linux when find_library fails
sdl = ctypes.CDLL(libname)

This should work, but as I don't have Windows to test this with, I rely on someone with Windows to test it, please.

python/settings_app/sdl_detect.py
evertvorster: Changed the hard-coded libSDL3.so.0 with ctypes.util.find_library like so:

Before (Linux-only)
sdl = ctypes.CDLL("libSDL3.so.0")

After (platform-agnostic)
libname = ctypes.util.find_library("SDL3")
if libname is None:
libname = "libSDL3.so.0" # fallback for Linux when find_library fails
sdl = ctypes.CDLL(libname)

This should work, but as I don't have Windows to test this with, I rely on someone with Windows to test it, please.

SGD1953:
Looks good to me, apart from the comments on SDL2 compatibility.

evertvorster:
> Looks good to me, apart from the comments on SDL2 compatibility.

Leaving out SDL2 here is intentional. The code base is moving away from it, and I'd like to move forward with clean code, rather than try to clean it up after the fact when SDL2 is not supported anymore. There is a separate PR for the last branch where SDL2 is still supported.

python/settings_app/sdl_detect.py
royfalk: Huh. I'm actually testing this on my MacBook, so nothing worked for me.
Claude recommended I replace the line with:
`import subprocess
prefix = subprocess.check_output(["brew", "--prefix", "sdl3"], text=True).strip()
sdl = ctypes.CDLL(f"{prefix}/lib/libSDL3.dylib")`

royfalk:
> > Looks good to me, apart from the comments on SDL2 compatibility.
> 
> Leaving out SDL2 here is intentional. The code base is moving away from it, and I'd like to move forward with clean code, rather than try to clean it up after the fact when SDL2 is not supported anymore. There is a separate PR for the last branch where SDL2 is still supported.

Unfortunately, kivy still relies on SDL2. There are efforts to move it to SDL3 but stable is still compiled against sdl2.

python/settings_app/sdl_detect.py
```diff
@@ -0,0 +1,95 @@
...
+        ("w", ctypes.c_int),
+        ("h", ctypes.c_int),
+        ("pixel_density", ctypes.c_float),
+        ("refresh_rate", ctypes.c_float),
+        ("refresh_rate_numerator", ctypes.c_int),
+        ("refresh_rate_denominator", ctypes.c_int),
+        ("internal", ctypes.c_void_p),
+    ]
+
+def main():
```
royfalk: I'm generally unhappy with writing a standalone cmd line utility and then running it from another program. 

But, that's something we can clean in another PR.

python/settings_app/sdl_detect.py
```diff
@@ -0,0 +1,95 @@
...
+    # Resolve the SDL3 library name for the current platform (SDL3.dll on
+    # Windows, libSDL3.dylib on macOS, libSDL3.so on Linux) instead of
+    # hardcoding the Linux name, so this works cross-platform.
+    libname = ctypes.util.find_library("SDL3")
+    if libname is None:
```
royfalk: There should be an option for Mac.
First, above
`import subprocess`

Then add a Mac check:
```
if libname is None:
   if sys.platform == 'darwin':
      prefix = subprocess.check_output(["brew", "--prefix", "sdl3"], text=True).strip()
    libname = f"{prefix}/lib/libSDL3.dylib"
   else:
   ...
```

royfalk (changes requested):
Two general comments:
1. With the reinforcement from Daniel, I think we can start porting this to the engine itself.
2. sdl_detect does not provide the native resolution of the panel. I did not dig to see why or whether this is a regression.

evertvorster:
> > > Looks good to me, apart from the comments on SDL2 compatibility.
> > 
> > 
> > Leaving out SDL2 here is intentional. The code base is moving away from it, and I'd like to move forward with clean code, rather than try to clean it up after the fact when SDL2 is not supported anymore. There is a separate PR for the last branch where SDL2 is still supported.
> 
> Unfortunately, kivy still relies on SDL2. There are efforts to move it to SDL3 but stable is still compiled against sdl2.
Yes, and that caused an issue, as you can't be calling an SDL3 routine inside an SDL2 application, hence the extra file. 

We could make this purely SDL2 for now, but then would have to come back to this when we either port it to the game engine, or update kivy to SDL3. 

Also, thanks very much for testing on your mac, I'll use your feedback with whatever we decide the direction to be here. 

What's the cleanest path forward here in your opinion? Do SDL2 only for now, or something else?

Also, I'm toying with getting all the bindings in the config app now. I have a branch on Assets that is not a PR yet, until I have something that works on my machine. I know the engine does not read the bindings from there yet, so I'm taking some liberties with how bindings are laid out in the json.

evertvorster:
The code is now SDL2 only, to match Kivy. This made the changes much, much more simple, and I now rely on Kivy's code to properly detect the OS and libraries since it needs to do that to work in any case.

royfalk (approved):
OK. It seems to work well.
I did not test on Windows. 
If this is something that is required, I can probably do this next week.

stephengtuggy:
> OK. It seems to work well. I did not test on Windows. If this is something that is required, I can probably do this next week.

I think we'd better test this on Windows before we merge it. I might also be able to do this this coming week.

SGD1953:
@evertvorster could we please squash the commits into a single one to keep the git history a little less noisy?

stephengtuggy:
> @evertvorster could we please squash the commits into a single one to keep the git history a little less noisy?

@SGD1953 For Vega Strike, we prefer not to squash commits or rebase in general. We made that decision several years ago. (Sorry!)

stephengtuggy:
> > @evertvorster could we please squash the commits into a single one to keep the git history a little less noisy?
> 
> @SGD1953 For Vega Strike, we prefer not to squash commits or rebase in general. We made that decision several years ago. (Sorry!)

@BenjamenMeyer we should probably make this part of our documentation too, huh?

SGD1953:
I can understand why not squashing them on merge, but when the author squashes them into more meaningful and compact sets?
I.e. here we have:
- implemented SDL3
- implemented hybrid
- removed SDL3 and kept SDL2
- some fixes

To me this is just "Migrated screen detection to use native SDL2" and unnecessarily spams the history.

evertvorster:
> I can understand why not squashing them on merge, but when the author squashes them into more meaningful and compact sets? I.e. here we have:
> 
> * implemented SDL3
> * implemented hybrid
> * removed SDL3 and kept SDL2
> * some fixes
> 
> To me this is just "Migrated screen detection to use native SDL2" and unnecessarily spams the history.

I see your point, and will change my workflow to have more sensible commits. Currently I was iterating by pushing up the code to github, and then pulling and building with an AUR package. I'll make a custom package that uses the local repo. That way I can iterate in peace, and only push up when I'm happy with the results. 

I guess that would make everyone happy?

That said, this should not stand in the way of approving *this* PR.

evertvorster:
> > @evertvorster could we please squash the commits into a single one to keep the git history a little less noisy?
> 
> @SGD1953 For Vega Strike, we prefer not to squash commits or rebase in general. We made that decision several years ago. (Sorry!)

I did do a couple of forced pushes when I went off totally in the wrong track, and then back-tracked. Rest assured that my reasons for doing this are sound. However, I will change my workflows to not have to do this in the future, and only commits will be made that I am happy with at the time.

SGD1953:
Force-pushes on your development branch are not an issue to me, they are a normal part of history hygiene, unlike for release branches or even master of course.

BenjamenMeyer:
> I can understand why not squashing them on merge, but when the author squashes them into more meaningful and compact sets? I.e. here we have:
> 
> * implemented SDL3
> * implemented hybrid
> * removed SDL3 and kept SDL2
> * some fixes
> 
> To me this is just "Migrated screen detection to use native SDL2" and unnecessarily spams the history.

@SGD1953 so the basic policy is not to squash/rebase something once shared as that can cause issues for others when they are trying to help you out or monitoring your changes. You are, of course, free to do anything locally or with branches you haven't shared. It also helps with detecting where and why changes come in - especially when using git-bisect to triage issues to their root cause.

don't worry about the history spam. That's really a non-issue.

> Force-pushes on your development branch are not an issue to me, they are a normal part of history hygiene, unlike for release branches or even master of course.

Generally yes, but once you share it and make it known then that can become and issue for anyone you are collaborating with. I'd probably make the cut-off when announcing a branch to people or if you go directly to filing a PR. If you haven't told people about a branch then certainly do what you will with it.

> I did do a couple of forced pushes when I went off totally in the wrong track, and then back-tracked. Rest assured that my reasons for doing this are sound. However, I will change my workflows to not have to do this in the future, and only commits will be made that I am happy with at the time.

There are certainly some exceptions to the general rules. If you do have to do that on a shared branch then its a matter of communication to folks that may be effected. Most of the exceptions are probably around security (e.g accidentally committed an API Key, password, etc) or if there was something extremely messed up and needing to do a `git reset` to move backwards in history to resolve the branch. However they should certainly be rare cases.

Hope that makes sense.

python/settings_app/sdl_detect.py
evertvorster: The SDL3 part is gone, as is the file needed to run it. The utility is fully SDL2 now, because Kivy is SDL2, and the detection routines are contained in osutil

python/settings_app/sdl_detect.py
evertvorster: Your concern has been noted, and acted upon. The file is gone, and the detection utility is now SDL2 only, re-using the same libraries Kivy uses. Please check to see if this PR now detects your monitor resolution, please

python/settings_app/sdl_detect.py
evertvorster: This is gone now, re-using the SDL2 init and libraries of Kivy. So, it should just work for you now.

evertvorster:
@BenjamenMeyer I see you requested code changes, and the changes you requested that I can see are in a file that is gone now. It has been tested to work on Linux and Mac, so we are just waiting for a Windows test before merging?

BenjamenMeyer:
> @BenjamenMeyer I see you requested code changes, and the changes you requested that I can see are in a file that is gone now. It has been tested to work on Linux and Mac, so we are just waiting for a Windows test before merging?

i'm okay with it being merged; @stephengtuggy might be able to test the Windows side.

royfalk:
I'll try and test this on Mac and Windows this weekend, and by weekend I mean Sunday.
Sunday is a weekday for us and my kid goes to summer school, so I have some free time. :)

PR #261, opened 2026-08-01
Sdl2 sdl3 fallback for 0.10.x

evertvorster:

Please answer the following:

Code Changes:

This only affects the settings app. 

Issues:
- Settings app mis-detects monitor resolutions on Wayland. 

Purpose:
* Avoid screeninfo, and use the same SDL that the engine use. 
- What release is this for?
* 0.10.x

### DIFF
--- a/python/settings_app/os_utils.py
+++ b/python/settings_app/os_utils.py
@@ -1,5 +1,5 @@
+import ctypes
 import os
-import screeninfo
 import sys
 
 import app_config as ac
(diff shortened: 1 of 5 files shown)


SGD1953 (approved):
Great step forward from the previous version, did you test the SDL2 path?

evertvorster:
> Great step forward from the previous version, did you test the SDL2 path?

Yes, **within limits** ie, I'm not about to uninstall SDL3 from my system to check the code in place and fallbacks. However, that part of the code, when run in isolation detects the SDL2 that I have installed, and returns the same artifacts as the SDL3 path.

PR #262, opened 2026-08-01
Capture all bindings in the config file, and modify them with the settings app. 

evertvorster:

Please answer the following:

Code Changes:

The intent of this PR is to have a config file in which each of the bindings are defined, and this config file is the **primary artifact**.

The shape of the bindings dictated how they should look in the JSON file. Each action can be bound to multiple input events from multiple input devices. Yes, this doubled the size of an already impressively long file, but hopefully a human would never have to read this again. 

What this addition to the config file enables, is to move the engine from reading the bindings from the old xml based config file over to the json config file, and should get us almost if not all the way to finally retiring the old settings app and configuration file. 

To bootstrap this process, I mined the xml file for all the bindings for all input devices and populated this config file. This means that hopefully it won't grow much longer. 

The settings app was then modified to enable these bindings to be modified. 
The Keyboard tab was replaced with an Action tab. 
Each action is read from the config file, and what binds to it can be modified. Each binding occupies one line. Lines can be modified, cleared out or more added, depending on which bindings to a certain action is required or desired. 

To handle binding to joystick axes, a section was added to the Controls tab. When a joystick is connected, moving the joystick enables the axis for display, and you can select which axis to bind to which joystick axis. 

Multiplayer bindings were not ported. Analog HAT switches were not ported, as I don't have access to such hardware, and considering what a HAT switch actually is, are likely to be all digital. If someone ever requires support for analog HAT switches, we can add support for it at that time. 

So, this PR is to be viewed as a **proof of concept** and an attempt to visualize how we are to deal with bindings. 

Comments and questions are welcome. 

Quite a few changes were made to the settings app. While not the main aim of this PR, it does add some nice functionality that did not exist in VS before. 

Here is a report on the code changes made to guide anyone brave enough to review this PR:
[pr-review-guide-actions-config.md](https://github.com/user-attachments/files/30712851/pr-review-guide-actions-config.md)

Issues:
- No existing way of adjusting bindings with a GUI
- Game engine still reads bindings from a xml based config file, which has to be hand-edited, and contains a LOT of unrelated stuff that has already been ported over. 

Purpose:
- - Open the door to finally moving away from the old config file
- - Showcase a way of doing bindings to actions. 
- What release is this for?
- - Git master. 
- - No. And it will require the engine to be modifed to actually read the bindings from this config file before any of this is actually useful.


config.json
```diff
@@ -1112,6 +985,711 @@
     "keyboard": {
         "enable_unicode": true
     },
+    "actions": {
```
stephengtuggy, line 988: I like it! What about naming the section "bindings" though, instead of "actions"? I guess I could go either way on that.

evertvorster: Good question. I went with actions, as that was the starting point. The game defines a list of actions, and that is what populates the config. The bindings are secondary. However, it would be a tiny change if you have strong feelings about it.

python/settings_app/graphics_factory/label_control_pair.py
```diff
@@ -13,22 +15,55 @@
...
+    0x200, which implies SDL_INIT_EVENTS) - Kivy's window provider does NOT
+    initialize the joystick subsystem, so without this SDL_NumJoysticks always
+    returns 0 and no axes can be enumerated. Shared by the joystick
+    enumeration and hat-polling paths. Returns the ctypes handle, or None if
+    SDL2 can't be loaded.
+    """
+    import ctypes
+    import ctypes.util
+    try:
+        sdl = ctypes.CDLL(ctypes.util.find_library("SDL2") or "libSDL2-2.0.so.0")
```
BenjamenMeyer, line 31: since this maps SDL2 it should be against 0.10.x branch instead of master; now that we have a 0.10.x branch

config.json
royfalk: +1 for bindings as well.

royfalk, line 988: Also, would prefer to see this as a separate JSON file, given its size.

python/settings_app/graphics_factory/label_control_pair.py
```diff
@@ -13,22 +15,55 @@
 from kivy.core.window import Window
 
 
+def _load_sdl():
```
royfalk, line 18: I'd prefer this in a separate file. Something like sdl_utils.py and either in root or in a utils folder.

python/settings_app/graphics_factory/label_control_pair.py
```diff
@@ -13,22 +15,55 @@
...
+    sdl.SDL_JoystickNumAxes.restype = ctypes.c_int
+    return sdl
+
+
 class AbstractLeafGui(BoxLayout):
     def __init__(self, parent: BoxLayout, leaf: gc.ConfigLeaf, title = None, tooltip_text = None):
         super().__init__(orientation='horizontal', height=70, size_hint_y=None)
         self.leaf = leaf
-        title  = title or f"{key_utils.format_key(leaf.key)}:"
+        if title is None:
```
royfalk, line 53: Why the change?

python/settings_app/graphics_factory/label_control_pair.py
```diff
@@ -13,22 +15,55 @@
...
 
-        if tooltip_text:
-            label = TooltipIcon(text=title, tooltip_text=tooltip_text, valign='middle', halign="left")
-            label.bind(size=self.update_text_size)
-            self.add_widget(label)
-        else:
-            label = Label(text=title, valign='middle', halign="left", height = 70)
-            label.bind(size=self.update_text_size)
-            self.add_widget(label)
+        if title:
```
royfalk, line 58: What happens `if title == None`? Why check this?
We just did `if title == None` above.

python/settings_app/graphics_factory/label_control_pair.py
```diff
@@ -172,41 +207,285 @@ def on_change(self, instance, new_text):
...
-            # TODO: validate?
-            modifier = modifiers[0]
-            self.key_leaf.set(codepoint)
-            self.modifier_leaf.set(modifier)
-            text = f"{codepoint} ({modifier})"
-            self.keystroke_label.text = text
-            if isinstance(keycode, int):
-                print(f"Key pressed: {keycode}, {scancode}, {codepoint}, Modifiers: {modifiers}")
-            
+class CaptureBindingButton(AbstractLeafGui):
```
royfalk, line 210: Split to its own file. Too long.

python/settings_app/graphics_factory/label_control_pair.py
```diff
@@ -172,41 +207,285 @@ def on_change(self, instance, new_text):
...
+        self.device = device
+        self.on_capture = on_capture
+
+        self.button = Button(text=self.format_binding(), size_hint=(0.8, None), height=35,
+                             pos_hint={'center_y': 0.5})
+        self.add_widget(self.button)
+        self.button.bind(on_press=self.on_click)
+
+    @staticmethod
+    def device_from_binding(binding: dict) -> str:
```
royfalk, line 238: Document the why of this function. I assume it's a helper of some kind. Maybe to process our input?

python/settings_app/graphics_factory/label_control_pair.py
```diff
@@ -172,41 +207,285 @@ def on_change(self, instance, new_text):
...
+        self.button.text = self.format_binding()
+        if self.on_capture:
+            self.on_capture(device, new_binding)
+
+
+class BindingCaptureDialog(ModalView):
+    """Modal capture dialog. While open, the overlay swallows every touch,
+    so the rest of the app is inert. Any key press (including Esc), mouse
+    click, or joystick button is captured and shown; the user then chooses
+    Accept / Retry / Cancel. On accept, on_capture(device, binding_dict) is
```
royfalk, line 294: What does retry do? I can always press again before Accept or Cancel.

python/settings_app/graphics_factory/label_control_pair.py
```diff
@@ -172,41 +207,285 @@ def on_change(self, instance, new_text):
...
+
+    def _on_key(self, window, keycode, scancode, codepoint, modifiers):
+        # Return True to consume the event: window_sdl2 dispatches on_keyboard
+        # (which quits the app on Esc via exit_on_escape) only when on_key_down
+        # returns falsy. Consuming every key here keeps the app alive while
+        # the capture dialog is open.
+        import key_utils
+        if keycode in key_utils.MODIFIER_KEYCODES:
+            return True
+        engine_key = key_utils.keycode_to_engine_name(keycode, codepoint)
```
royfalk, line 360: Not sure I would use engine here. Maybe game? @stephengtuggy ?

python/settings_app/graphics_factory/label_control_pair.py
```diff
@@ -172,41 +207,285 @@ def on_change(self, instance, new_text):
...
+                    elif v & 4:  # SDL_HAT_DOWN
+                        vy = -1
+                    if v & 2:   # SDL_HAT_RIGHT
+                        vx = 1
+                    elif v & 8:  # SDL_HAT_LEFT
+                        vx = -1
+                    self._on_hat_value(i, h, (vx, vy))
+                self._hat_state[key] = v
+
+    def _on_hat_value(self, stickid, hatid, value):
```
royfalk, line 421: There was a complain we did not support joysticks with 50 buttons or some such a while back. 

I assume this code parses the various buttons. Are these enough? 

Didn't that guy move to SDL3 so he could have more buttons.

python/settings_app/graphics_factory/label_control_pair.py
```diff
@@ -172,41 +207,285 @@ def on_change(self, instance, new_text):
...
+
+    # --- flow ---
+
+    def _set_binding(self, binding):
+        self._binding = binding
+        self._captured.text = CaptureBindingButton.format_binding_for(
+            CaptureBindingButton.device_from_binding(binding), binding)
+        self._accept_btn.disabled = False
+
+    def _retry(self):
```
royfalk, line 463: This button looks redundant from a UI point of view. If you don't like the result, either click again or press cancel.

python/settings_app/graphics_factory/label_control_pair.py
```diff
@@ -227,8 +506,255 @@ def on_slider_change(self, instance, value):
         print(f"Slider changed from {self.leaf.value} to {value}")
         self.leaf.set(value)
 
-        
 
+class LiveAxisSlider(BoxLayout):
```
royfalk, line 510: Also split to a separate file. Did not review this as I did not connect joystick. Not sure what it does.

Why are we setting a deadband. Is that where the subtle move of the joystick doesn't move the ship?

python/settings_app/tabs/bindings.py
```diff
@@ -1,40 +1,184 @@
...
 import game_config as gc
-from graphics_factory.label_control_pair import CaptureKeyStrokePair
+from graphics_factory.label_control_pair import CaptureBindingButton
 from graphics_factory.divider import DividerLine
 
+# Device types (display name -> config key)
+DEVICE_CHOICES = [('Key', 'keyboard'), ('Mouse', 'mouse'), ('Joy', 'joystick'), ('Hat', 'hat')]
 
 
+class BindingRow(BoxLayout):
```
royfalk, line 14: I did not review this, but I did play test it. 
Looks great. A real improvement.

python/settings_app/tabs/controls.py
```diff
@@ -1,11 +1,93 @@
...
+                                                 AxisExplorer)
+
+
+def _set_leaf(gc_obj, path, value):
+    leaf = gc_obj.get_object(path)
+    if leaf is not None:
+        leaf.set(value)
+
+
+class FlightControlSelector(BoxLayout):
```
royfalk, line 20: From a usability point, I did not understand the GUI.
I'd expect something like:
Keyboard
Keyboard and Mouse
Keyboard and Joystick

royfalk, line 20: Also, maybe move this class to its own in graphics factory?

python/settings_app/tabs/controls.py
```diff
@@ -15,49 +97,140 @@ def __init__(self, **kwargs):
...
+        # ScrollView handles any overflow.
+        self.options_container = BoxLayout(orientation="vertical", size_hint_y=None,
+                                           height=0)
+        self.options_container.bind(
+            minimum_height=self.options_container.setter('height'))
+        mode_boxes = {"Keyboard": [], "Mouse": [self.mouse_box],
+                      "Joystick": [self.axes_box]}
+        FlightControlSelector(parent=content, options_container=self.options_container,
+                              mode_boxes=mode_boxes)
+
```
royfalk, line 133: This would be redundant if we follow my suggestion.
Also, note that when clicking on the drop down list you get:
Keyboard
Keyboard
Mouse
Joystick

With **two** labels of keyboard is always enabled. I was actually confused by this in the beginning.

royfalk (changes requested):
That's really great work. 
You nailed it with the design for the key binding. 

I have a ton of minor comments. None of them are critical, so push back if something breaks the AI and we can polish them in a separate commit.

SGD1953:
First of all, thank you for this contribution, it is very much needed!

The way things are presented is not super-intuitive to me and not how other games would handle this. We don't need to do what everyone else does but I'd argue games are used to certain de-facto standards. Those are not blockers to me as moving away from XML and entirely towards JSON is a massive win and there is no way to set keybindings today.

1. Separate Keyboard, Joystick and Mouse into separate tabs - this is more aligned with industry best-practices and should eliminate the need to have multiple bindings for the same key in one list which is very hard to absorb
2. We should introduces groups (collapsible) and more human readable names/labels (like the tables in the manual - Camera Control - Navigation - Communication etc.)

Here are my play-testing findings, will update when I find more

- Double key bindings possible - I could have "z" bound to two actions (did not test what happens in-game)
- Swapped z/y -> works in-game
- Changed full throttle to "ß" -> works in settings-app (and seems to be properly saved to `\u00df`), does not work in-game (Error: `No such special key: ß`)

python/settings_app/key_utils.py
```diff
@@ -1,5 +1,117 @@
...
+        engine_name = KIVY_TO_ENGINE.get(kivy_name)
+        if engine_name:
+            return engine_name
+    if codepoint:
+        return codepoint
+    return None
+
+
+# Friendly display names for engine key names.
+ENGINE_KEY_DISPLAY = {
```
SGD1953, line 87: Nice!

SGD1953:
When I connect my controller, the ship keeps rolling right in flight.

Then I found that I can switch to Joystick in the Controls section and found the axis settings I was looking for. Switched things around so that the sticks are properly mapped - now it works (need to properly map keys though).

```
1 joysticks were found.

The names of the joysticks are:
    Xbox One S Controller
```

My USB cable might be a bit flaky, after the gamepad disconnects it fails to reconnect:
```
Joystick removed: slot 0
Joystick added but failed to open: slot 0
```

evertvorster:
Thanks for having a go at it, your feedback is very valuable, as it shapes where I am going with the settings app that is included in the engine. 

Just off-hand, you can bind any input to any action, so you have a list of actions, and binds to those. Would you like to see the same complete list again for mouse, and then again for joystick? 

No, the current thing works, but I agree that the list is very long, and that the actions can be grouped. 

The game does not do well with flaky joysticks.. Oh, while there is no calibration per se, there is an "Auto deadzone" button that picks a deadzone that is just slightly larger than your joystick's center position. 

I will have to check what happens when one input is bound to two actions. I'm not sure if its something I want to mess with, so I'll leave it in whichever state the engine currently supports. 

The overarching idea is to get to a sane config app in the game, and then hand it over to you guys if you want to make structural changes.

SGD1953:
> Would you like to see the same complete list again for mouse, and then again for joystick? 

No, I would want it to be filtered by device.

> No, the current thing works, but I agree that the list is very long, and that the actions can be grouped. 

Great!

> The game does not do well with flaky joysticks.. Oh, while there is no calibration per se, there is an "Auto deadzone" button that picks a deadzone that is just slightly larger than your joystick's center position. 

Deadzone is not the issue, I saw that, but once it disconnects it disconnects for good, there might be an issue with it reclaiming that same slot. You could try unplugging yours to reproduce.

The other thing is, there are no good gamepad defaults, if we could detect joystick vs. gamepad and have defaults for both that would be amazing.

> I will have to check what happens when one input is bound to two actions. I'm not sure if its something I want to mess with, so I'll leave it in whichever state the engine currently supports. 

I don't think we need to fix it in the engine, but we could reset the binding which uses the same key in the settings app. Or at least warn when there are duplicates.

> The overarching idea is to get to a sane config app in the game, and then hand it over to you guys if you want to make structural changes. 

Yes, once my other PRs are merged I'll have a look at how easy it is to add a settings room to the main menu.

evertvorster:
> > Would you like to see the same complete list again for mouse, and then again for joystick?
> 
> No, I would want it to be filtered by device.
> 

Once again.. very slowly for the people at the back:
The engine supports binding any input to any action. 

There is no "Filter by device". There cannot be, you'll get the full list of available actions, all the time. 

Lets say you start with your input device. Detecting what the specific inputs the device has available is easy, mapping them to actions is a different story. For each input, you would **have** to display all the possible options, 143 different options to be exact.  Good luck sorting through that. 

***It does not work, it is impractical*** 
I have tried, and failed. Many times. You are welcome to try too. 


The way it DOES work, is that you start with the list of actions. 
These actions have a default keyboard binding, which can be changed. 

If you see an action that you would like to also bind to an input device, you press a button, perform the action, and confirm it. Done. 

So, in the current way, you only have to go through the list once. 

The only improvement to be made is splitting the list into categories, so you have views, flight control, communication and so on to make finding the right action easier. 

In the future, I could add some *display* filters so you can see what your joystick has been bound to, or your mouse, but this will still operate on the same list of actions. I'm also considering some buttons for un-binding everything from a device, basically resetting it, and also if needed some conflict detection, say if you have the same input mapped to two actions. I don't even know if the engine supports that, and what will happen in this case. 

My initial concern is to make it work. We can make it pretty later.

evertvorster:
> > The overarching idea is to get to a sane config app in the game, and then hand it over to you guys if you want to make structural changes.
> 
> Yes, once my other PRs are merged I'll have a look at how easy it is to add a settings room to the main menu.

That might be a while. 

However, we'll be working on the same concept, we might as well collaborate. 

When I get something that works, I can invite you to do a review. However, this might be a while, as having a config room (I have it as a screen utility that pops up with a hotkey at the moment)  totally depends on actually being able to make configuration changes, which in turn depends on sane config files and mappings into the engine.

SGD1953:
> Once again.. very slowly for the people at the back:
> The engine supports binding any input to any action.
>
> There is no "Filter by device". There cannot be, you'll get the full list of available actions, all the time.
>
> Lets say you start with your input device. Detecting what the specific inputs the device has available is easy, mapping them to actions is a different story. For each input, you would have to display all the possible options, 143 different options to be exact. Good luck sorting through that.

Let me try again, while the engine does support this, it does not mean that the settings app has to **present** it this way. You could for instance turn it around and visualize a joystick or a controller and then allow to attach an action to the buttons, we can also be opinionated on the list of actions to show and hide the rest in an advanced section. There are **a lot** of options here.
 
> It does not work, it is impractical
> I have tried, and failed. Many times. You are welcome to try too.
>
> The way it DOES work, is that you start with the list of actions.

I have to respectfully disagree here, just because you could not get it to work does not mean it does not work.

I'd like to highlight that the feedback provided here is just that - feedback. You can take it or leave it, it would be great if we could stay open-minded along the process as I have see quite some games which support keyboard, mouse and joystick and none of them ever presented it this way. Thus claiming this to be the only way is a bit of a stretch. I completely agree that it would be a major win to discontinue XML and if the price is to temporarily accept and not so great settings UI that's a tradeoff we can make.

> However, we'll be working on the same concept, we might as well collaborate.

100%, I think there is a good overlap here, I spent quite some time working on the GUIs/bases and you know a lot of how the settings/mappings work. That's also the reason for the feedback on the should-be state as apart from merging this or not it would be great if we were all aligned how how the end state of this would look like.

BenjamenMeyer (changes requested):
@stephengtuggy created the 0.10.x branch in this repo; can we please get this based against that branch instead of `master`? Then we can also get an SDL3 version against master.

python/settings_app/graphics_factory/label_control_pair.py
stephengtuggy: Yes, that's what a deadband is. So that your joystick isn't hypersensitive to every move you make with it.

python/settings_app/graphics_factory/label_control_pair.py
stephengtuggy: I think it's fine, personally. We could call it `game_engine_key`, but that's probably longer than necessary.

python/settings_app/graphics_factory/label_control_pair.py
stephengtuggy: I just rebased this branch against 0.10.x.

PR #263, opened 2026-08-02
Main menu rework

SGD1953:
This is a set of changes which create a nice main menu with hover states for buttons. No version number for the master branch.
vs_main_menu


bases/main_menu.py
```diff
@@ -278,46 +278,45 @@ def onSkip(self, button, params):
...
 sprite = {
-        '*':None,
-        'down' : ( 'interfaces/main_menu/credits_button_pressed.spr', sprite_loc ) }
+        '*': ( 'interfaces/main_menu/credits_button.spr', sprite_loc ),
+        'hot' : ( 'interfaces/main_menu/credits_button_hover.spr', sprite_loc ) }
 GUI.GUIRoomButton(guiroom, credits_guiroom, 'XXXShow Credits','Show_Credits',sprite,sprite_loc,clickHandler=enterCredits)
 
-# Link
-sprite_loc = GUI.GUIRect(48,440,150,32,"pixel",(1280,1024))
+# Into button
```
stephengtuggy:
```suggestion
# Intro button
```

stephengtuggy (changes requested):
Found one typo. Other than that, it looks good, although I haven't play-tested it yet.

stephengtuggy:
@evertvorster in case you're interested and get a chance to review this.

bases/main_menu.py
SGD1953: Typo fixed

evertvorster:
Quick question before I dive in... Does this PR depend on the PR that broke being able to type in the name field for saved games?

evertvorster (changes requested):
Your branch, as it stands, has something very wrong with it. 
Against git master version of the game engine:
1. Loads up, NO buttons. If you added new sprites and such, please ensure that they are actually installed when you run make install. I play from installed files, not the source repo. 

2. After right and left clicking a bit, I was able to launch a new campaign. 

3. Clicking on the base computer crashes the game. <- On Engine git master. 
On https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1684 (which this was designed against) The base computer runs, and works normally. 

For future reference, if you build your Assets against a specific engine, these need to be reviewed together, and approved together. Also, you need to mention if they rely on both being there at the same time... Ie: be very specific about the relationship between them. 

This makes my job a lot easier, and I can test your stuff quicker.

I did poke at the no buttons problem. I see the .spr files have been installed, but they point to .ani files which are nowhere to be found.

SGD1953:
> Your branch, as it stands, has something very wrong with it. Against git master version of the game engine:
>

Thank you for taking the time to test this, it is very much appreciated! The timing of it was a bit unfortunate I'm afraid.
 
>     1. Loads up, NO buttons. If you added new sprites and such, please ensure that they are actually installed when you run make install. I play from installed files, not the source repo.

This is odd, I just retested it (after merging master to be fair) and it works just fine.

> 
>     2. After right and left clicking a bit, I was able to launch a new campaign.
> 

This is already too far, this PR only touches the main menu.

>     3. Clicking on the base computer crashes the game. <- On Engine git master.
>        On [vegastrike/Vega-Strike-Engine-Source#1684](https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1684) (which this was designed against) The base computer runs, and works normally.

This is because other changes had been merged in the meantime and the controls.json which engine master requires was missing here.
 
> 
> For future reference, if you build your Assets against a specific engine, these need to be reviewed together, and approved together. Also, you need to mention if they rely on both being there at the same time... Ie: be very specific about the relationship between them.
> 

I don't think there was an actual dependency, the issue was this feature branch being out of date.

> This makes my job a lot easier, and I can test your stuff quicker.

100%, my apologies again.

> I did poke at the no buttons problem. I see the .spr files have been installed, but they point to .ani files which are nowhere to be found.

That is "normal" for all I know.

stephengtuggy:
What is the status of this PR?

SGD1953:
@stephengtuggy don't merge it for now, I just tested it again and the buttons are gone. Will circle back once resolved

PR #267, opened 2026-08-08
Compressed cargo images with bc3 and improved consoles and holovids

SGD1953:
This PR uses a higher quality compression (DXT5) for cargo images (as it should have in the first place) and includes the improved visuals for consoles and holovids


royfalk:
How do I test this?

SGD1953:
If you locally pull the branch you can look at the images via for example gimp or another viewer which supports DDS. GitHub might even show then once we use the proper DDS extension. Not sure where the cargo images are referenced and how much work that would be.

BenjamenMeyer (approved):
@royfalk I pull the branch locally and then just look at the folders with KDE's Dolphin - it renders stuff well enough I can see. Sometimes I've swapped back and forth between the target branch (master) and source to see the differences. Works fairly well for most images; but I'd love it we could use extensions that GH would just render for us.

SGD1953:
Are we good to merge here?

royfalk:
Well. That's going to be a bit harder in a hotel in Crete. 😁

stephengtuggy:
@SGD1953 can we get this backported to 0.10.x as well?

PR #268, opened 2026-08-09
Split config concepts of a plan

evertvorster:
**Read this first**

What you are looking at in this PR is a suggestion for the shape of data in a four-way split of the config.json. 

You'll see a deeper level of bindings, and the concept of presets. 

Everything that was settable with the original settings app is available to set. 

In broad terms, the shape here is as such:
**config.json** = Typically what a user would set. 
**bindings.json** = Separated from config.json to allow for language specific bindings in the future. 
**theme.json** = Pure assets file, where the colors of things are defined. The xml file has the info in this way, so we just re-create it. Once merged, we can massage the data into new ways. 
**engine.json** = The user can't set any of this. This is what people that make different universes would set. (We can do little fine tunes here, but basically this is where the game engine is configured.) This is also where there is a bunch of presets, and we'll probably add in game difficulty here as well. 

What I want you to do with this PR:

Have a look at the data structures in each file. Decide if it is appropriate for the data stored in that file. 

Decide whether this split of the config.json makes sense.  Some things might move between the files later based on practicality, but this here is broad strokes. 

Don't worry about the specifics of a single setting or so, there will be changes by necessity.

If you have any comments, please feel free to add them here, and we can massage the layout till it makes sense. 

Once I have agreement that this is the way to go, I'll proceed to the next step. Rest assured, this won't be merged unless everyone is happy with the shape of the config files, and there is an engine PR that at least reads the config variables that moved out of the config.json into the split files.  This is the first hurdle we have to cross, as changing the format of a file is much harder once we start building on it. 

This way, I hope to keep the subsequent PRs small. This one is massive, as there is a lot of data that was mined out of the old xml config, and huge chunks of data was also moved around. 

TL;DR: This is my proposed layout for a split config file setup, pre-filled so it won't grow much larger.


bindings.json
```diff
@@ -0,0 +1,1226 @@
+{
```
royfalk, line 1: Good

config.json
```diff
@@ -1,1493 +1,139 @@
 {
-    "settings_app": {
```
royfalk, line 2: Not sure why this branch and the one below (advanced) were removed.

Well, the advanced branch should gradually be filled with the stuff we think users aren't usually interested in.

Or did we have an advanced branch for each section?

engine.json
```diff
@@ -0,0 +1,1824 @@
+{
```
royfalk, line 1: OK. So all the "advanced" settings moved here. If I'm an ordinary user and I want to edit something, can I do it?

IIRC, this is not the case for people who don't compile their game. It gets installed somewhere read only.

Also, this potentially breaks the engine, as we don't generate a configuration.cpp/h from this.

theme.json
```diff
@@ -0,0 +1,242 @@
...
+        0.3,
+        1.0
+      ],
+      "destination_system": [
+        1.0,
+        0.77,
+        0.3,
+        1.0
+      ],
+      "slelection_system": [
```
royfalk, line 16: typo and also maybe should be selected_system?!

royfalk (commented):
Biggest issues I see are:
- moving stuff that is in the config.json will break configuration.cpp
- can't edit the engine file without permissions.

Neither is a complete show stopper. We can move from a single config to multiple config files and still have a unified config in the engine.

This is definitely one place where I'd want input from @stephengtuggy and @BenjamenMeyer. 

If you don't want to wait, I'd start with the low hanging fruit.
- Take out the binding section to its own json. Create a corresponding C++ class to parse and store it.
- Modify engine to read the bindings from there.
- Modify the settings app to write the bindings to .vegastrike.

config.json
evertvorster: Yeah, advanced has to go away. The normal player can set a few things that are obvious, but physics and that sort of stuff was moved out of config.json on purpose. This is where Asset creators play. They set up the rules of the universe the players play in. :)

engine.json
evertvorster: Not without root. Ordinary users don't get to mess with the internals.

evertvorster:
> Biggest issues I see are:
> 
> * moving stuff that is in the config.json will break configuration.cpp
> * can't edit the engine file without permissions.
> 
> Neither is a complete show stopper. We can move from a single config to multiple config files and still have a unified config in the engine.
> 
> This is definitely one place where I'd want input from @stephengtuggy and @BenjamenMeyer.
> 
> If you don't want to wait, I'd start with the low hanging fruit.
> 
> * Take out the binding section to its own json. Create a corresponding C++ class to parse and store it.
> * Modify engine to read the bindings from there.
> * Modify the settings app to write the bindings to .vegastrike.

Yeah, not my plan for the first step. 
I want all that engine configuration (advanced) stuff out of the config.json. So, that will be my first step, and yes, I'll have to get the engine to read it. 

The settings app might also need small changes, but then that's it. 

So, my plan for the next step is to do just that, and then get this and that PR merged. 


Once merged, I'll slowly move the engine over to reading from the new config files, and wean it off the xml file. I won't be needing changes to Assets for a while, and should be able to make incremental PRs on engine only.

config.json
stephengtuggy: > Yeah, advanced has to go away. The normal player can set a few things that are obvious, but physics and that sort of stuff was moved out of config.json on purpose. This is where Asset creators play. They set up the rules of the universe the players play in. :)

I disagree with this decision. It is not in keeping with long-standing Vega Strike traditions. Just about anyone is supposed to be able to change just about any setting. At least from what I've seen.

Please keep the advanced branch in place. It can "gradually be filled with the stuff we think users aren't usually interested in," like Roy said.

engine.json
stephengtuggy: > Not without root. Ordinary users don't get to mess with the internals.

I disagree with this decision as well. @BenjamenMeyer what do you think?

bindings.json
stephengtuggy: Yeah, I like separating out the bindings. I like it a lot.

engine.json
stephengtuggy: I would be fine with moving these settings to a separate file called something like "advanced_config.json", and letting users edit it. I think that would make a lot of sense.

config.json
stephengtuggy: OK. So the advanced branch doesn't absolutely have to stay in this location. I would be fine with moving it to a separate file, called "advanced_config.json" or something like that. Just make sure that users can edit it. That's all I ask.

engine.json
evertvorster: It is all in one file already. Engine.json.

config.json
evertvorster: OK, how about I treat the four JSON files in a similar way: If there is no file of an equivalent name in your .vegastrike directory, load the default. If there is a file, merge it with the file in the .vegastrike directory just like we do for config.json. 

A person who is interested in modifying the physics of the engine can do a one time: sudo cp /usr/share/vegastrike/engine.json ~/.vegastrike and then modify to hearts content. 

For someone that just wants to sit down and play, we don't want to show a bewildering number of options that when fiddled with breaks the game. 

Now, if you still have strong feelings about this, once I have done the settings app, feel free to extend the functionality to also modify the physics and so on.

stephengtuggy: > OK, how about I treat the four JSON files in a similar way: If there is no file of an equivalent name in your .vegastrike directory, load the default. If there is a file, merge it with the file in the .vegastrike directory just like we do for config.json.
> 
> A person who is interested in modifying the physics of the engine can do a one time: sudo cp /usr/share/vegastrike/engine.json ~/.vegastrike and then modify to hearts content.
> 
> For someone that just wants to sit down and play, we don't want to show a bewildering number of options that when fiddled with breaks the game.
> 
> Now, if you still have strong feelings about this, once I have done the settings app, feel free to extend the functionality to also modify the physics and so on.

All right. I am OK with this.

stephengtuggy (approved):
LGTM. Let's wait until @BenjamenMeyer weighs in. If he is OK with this PR, then I think it can be merged, along with any needed changes to the engine to keep it from breaking.

Question: Should we add the new files to CMakeLists.txt so they get installed with the rest of the assets?

engine.json
BenjamenMeyer: > > Not without root. Ordinary users don't get to mess with the internals.
> 
> I disagree with this decision as well. @BenjamenMeyer what do you think?

Still have to read through a lot of this but to answer the specific...

It depends on the platform and where the file is installed.

Linux - yes, if we installed the file into /etc or /usr/share then yes it will only be modifiable by admin. user settings would need to go into `~/.vegastrike`; though we probably should consider using `~/.config/vegastrike` instead so we're more standards compliant.

MacOS - users have an easy way to access root privileges so it'll be easy for them to modify unless it is on the DMG for the assets.

Windows - users can typically modify files in the Program Files and Program Data directories where this would be stored. So it's not so much a "user cannot do this" as a "user does not know to do this". The user settings would be in their AppDir location under their user directory, preferably the Roaming location but I think we're just using `%user%/.vegastrike` at the moment.

So IMHO we do have some things to work on as far as where stuff is stored so it's easier to find and easier to ensure it gets backed up properly, but that's separate from this aside from how permissions interact with things.

bindings.json
```diff
@@ -0,0 +1,1226 @@
...
+    },
+    "LockTargetKey": {
+      "keyboard": [
+        {
+          "key": "l",
+          "modifier": "none"
+        }
+      ]
+    },
+    "MisSelKey": {
```
BenjamenMeyer, line 511: we should at some point expand things so it's clear what they are. `MisSelKey` is either `Mission Select Key` or `Missile Select Key`... not saying you need to in this PR, but since you're in there it may make sense.

bindings.json
```diff
@@ -0,0 +1,1226 @@
...
+      "joystick": [
+        {
+          "joystick": 0,
+          "button": 8,
+          "modifier": "none"
+        }
+      ]
+    }
+  },
+  "axes": {
```
BenjamenMeyer, line 1200: is this supposed to be `axises`?  `axes` are weapons - 🪓 🪓  as opposed to spacial movement.

Also this should probably be like above:
```json
"axises": [
   {
      "source": { "deviceClass": "joystick", "index": 0 },
      "type": "throttle",
       "axis": 3,
       "inverse": false,
   },
   ....
]
```
however it works out....since we could have multiple joysticks + keyboard + mouse in play for providing axis input.

engine.json
```diff
@@ -0,0 +1,1824 @@
...
+        "font_antialias": "false"
+      },
+      "text_aa_vector": {
+        "high_quality_font": "false",
+        "high_quality_font_computer": "false",
+        "font_antialias": "true"
+      }
+    },
+    "Geometry": {
+      "GeomRetro": {
```
BenjamenMeyer: I think we can just call these `Retro`, `Low`, `Medium` - no need to have the redundant `Geom` prefix on them since they're all under the `Geometry` space.

engine.json
```diff
@@ -0,0 +1,1824 @@
...
+    "FactionTextures": {
+      "faction_tex_off": {
+        "faction_dependant_textures": "0"
+      },
+      "faction_tex_on": {
+        "faction_dependant_textures": "1"
+      }
+    },
+    "Textures": {
+      "TexRetro": {
```
BenjamenMeyer: same here  - `Retro`, `256`, `512`.... they're all under `Textures` already

engine.json
```diff
@@ -0,0 +1,1824 @@
...
+      "show_stardate": true,
+      "smart_targetting_key": true,
+      "smooth_lines": false,
+      "smooth_points": true,
+      "smooth_shade": true,
+      "sparkle_absolute_speed": 0.02,
+      "sparkle_engine_size_relative_to_ship": 0.1875,
+      "sparkle_rate": 5.0,
+      "specmap_with_reflection": true,
+      "splash_audio": "../music/aerapeace.ogg ../music/news1.ogg ../music/isopeace.ogg ../music/asteroids.ogg ../music/militarybase_p.ogg ../music/terranpeace.ogg ../music/industrialmixed.ogg ../music/snow.ogg ../music/terranpeace2.ogg ../music/terranbattle.ogg ../music/galacticbattle.ogg ../music/peace4.ogg ../music/aerathemebattle.ogg ../music/peace3.ogg ../music/battle5_p.ogg ../music/alienrockalectro.ogg ../music/cosmostation_crowded.ogg",
```
BenjamenMeyer: this should probably be a list of strings:
```json
"splash_audio": [
    "../music/aerapeace.ogg",
    "../music/news1.ogg",
    ...
]
```

engine.json
```diff
@@ -0,0 +1,1824 @@
...
+      "star_color_average": 0.6,
+      "star_color_cutoff": 0.1,
+      "star_color_increment": 100.0,
+      "star_color_power": 0.25,
+      "star_body_radius": 0.33,
+      "star_glow_radius": 1.33,
+      "star_lumin_scale": 0.001,
+      "star_minimum_color_value": 0.3,
+      "star_overlap": 1.0,
+      "star_shine": "shine.ani flare1.ani flare2.ani flare3.ani flare4.ani flare5.ani flare6.ani",
```
BenjamenMeyer: same here

BenjamenMeyer (commented):
I think overall this makes sense. It also gives us a chance to go back and change some things - using a list of strings instead of `file1 file2 file3` that then has to be parsed down to a list of strings; cleaning up redudant naming of keys; expanding keys to be more meaningful (we don't need to be concerned about string length; but should keep things reasonable and descriptive).

Not sure anything is blocking... @royfalk @stephengtuggy can block on these if you like or we can follow up with another PR; though it'd probably be good to just get it done in one PR.

BenjamenMeyer:
@evertvorster reasoning for closing? Just curious if you don't mind.

evertvorster:
> @evertvorster reasoning for closing? Just curious if you don't mind.

I moved some things between the files, and I did not want to give the impression that I would get approval for one set of changes, and then abuse that to get a different set of changes in.

This PR was to outline the general idea, and to get feedback to fine-tune the final proposal that I want to merge. That one is live now. :)

BenjamenMeyer:
@evertvorster thanks; I see you have #270 continuing this effort.

theme.json
```diff
@@ -0,0 +1,242 @@
+{
+  "colors": {
+    "nav": {
+      "current_system": [
```
SGD1953: As we are refactoring. I'd suggest something like

```
"destination_system": {
        r: 1.0,
        g: 0.77,
        b: 0.3,
        a: 1.0
      },
```

instead of arrays.

theme.json
```diff
@@ -0,0 +1,242 @@
...
+      "space_background": [
+        0.0,
+        0.0,
+        0.0,
+        1.0
+      ]
+    }
+  },
+  "graphics": {
+    "afterburner_color_red": 1.0,
```
SGD1953: afterburner and engine values should be moved to colors.

bindings.json
SGD1953: Same, great idea.

config.json
```diff
@@ -1,603 +1,144 @@
...
-        "Enslave": {"key": "e", "modifier": "ctrl"}
-    },
-
-    "mouse": {
-        "enabled": true,
-        "FireKey": {"button": 0, "modifier": "none"},
-        "inverse_x": false,
-        "inverse_y": false,
-        "comment": "move joystick and keyboard here. Move bindings down a level."
+    "preset": {
```
SGD1953: I like the reintroduction of presets, I assum `computer` is the name of the preset?

Not top priority, but I'D love to see a sweep through names and values to make them more logical. Examples:

```
"geometry": "GeomVeryHigh", -> "vertexCount": "VeryHigh", // or even a value
"textures": "Tex1024", -> "textureResolution": "1024",
```

engine.json
```diff
@@ -0,0 +1,1742 @@
...
+                "colordepth": "32",
+                "zfar": "600000",
+                "fullscreen": "true"
+            }
+        },
+        "Resolution": {
+            "512x384": {
+                "x_resolution": "512",
+                "y_resolution": "384",
+                "aspect": "1.333",
```
SGD1953: I feel we should calculate aspect, if we want to keep the possibility to specify it we can introduce `auto`.

engine.json
```diff
@@ -0,0 +1,1742 @@
...
+                "font_point": "10"
```
SGD1953: This is something we should calculate as well, it should be a function of screen height. Same as above, we could have "auto" as default so people can overwrite it.

engine.json
```diff
@@ -0,0 +1,1742 @@
...
+            },
+            "800x600": {
+                "x_resolution": "800",
+                "y_resolution": "600",
+                "aspect": "1.33",
+                "font_point": "10"
+            },
+            "1024x600": {
+                "base_max_width__betternot": "800",
```
SGD1953: I like this, as the sprites are 4:3 and they get stretched. I'm afraid it is currently not properly implemented. We could calculate this too however to simplify the config. The config could then just say `keepSpriteAspect` with `true/false`.

engine.json
```diff
@@ -0,0 +1,1742 @@
...
+            "jump_engaged": "jump",
+            "manifest": "vdu_a",
+            "missile_switch": "vdu_d",
+            "objective": "vdu_c",
+            "overload": "overload",
+            "overload_stopped": "overload_stopped",
+            "repair": "vdu_a",
+            "scanning": "vdu_c",
+            "shield": "vdu_d",
+            "sounds_extension_1": "*_female.wav",
```
SGD1953: This could be `soundExtensionLookup: ['*_female.wav', '*.wav']` which will make handling in the engine easier as well.

engine.json
```diff
@@ -0,0 +1,1742 @@
+{
+    "presets": {
```
SGD1953: Numbers and booleans in this file are all strings, I'd prefer they were what they actually are which won't require us to parse them.

bindings.json
royfalk: Merriam Webster disagrees with you.
https://www.merriam-webster.com/dictionary/axis

engine.json
royfalk: Just be sure to update the engine code reading this.

bindings.json
BenjamenMeyer: > Merriam Webster disagrees with you. https://www.merriam-webster.com/dictionary/axis

interesting... another oddity of the English language

PR #270, opened 2026-08-14
Config split shape

evertvorster:
Code Changes:

Issues:
**THIS PR DEPENDS ON https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1719**

- - Config.json file becoming too long, and contains stuff that never changes. 
- - This PR opens the door for easy bindings editing without touching other config files.

Purpose:
- -Split the config.json into four files, each with a distinct purpose. 
- What release is this for?
- - 0.11
- -0.11

The config app is also modified to operate on this split files system. It will never be able to modify the engine or colors files. However, one **can** port over a setting from physics at a later stage if that is what you really want to do. 

Once this PR is merged, I can start work on the engine, swapping over the reading from the xml vegastrike.config to these json files.


theme.json
```diff
@@ -0,0 +1,242 @@
...
+        0.0,
+        0.0,
+        1.0
+      ]
+    }
+  },
+  "graphics": {
+    "afterburner_color_red": 1.0,
+    "afterburner_color_green": 0.0,
+    "afterburner_color_blue": 0.0,
```
BenjamenMeyer, line 233: this section would be better if instead of using a prefix it just became another JSON key.
```json
"graphics": {
   "afterburner": {
      "color": {
          "red": 1.0,
          "green": 0.0,
          "blue": 0.0,
      }
   },
   "engine": {
      "color": {
          "red": 1.0,
          "green": 1.0,
          "blue": 1.0,
      }
   },
   "base": {
      "faction": {
         "color": {
              "darkness": 0.75
         }
      },
      "saturation": 1.0,
      "set_raster_text_color": true,
  }
}
```

theme.json
```diff
@@ -0,0 +1,242 @@
...
+        1.0,
+        1.0
+      ],
+      "target": [
+        1.0,
+        0.0,
+        1.0,
+        1.0
+      ],
+      "planet": [
```
BenjamenMeyer, line 174: This could be associated a little better.
```json
{
   "planet": {
         "default": {...},
         "ambient": {...},
         "matte_ambient": {...},
         ...
   }
}
```

theme.json
```diff
@@ -0,0 +1,242 @@
...
+      ],
+      "engine": [
+        1.0,
+        1.0,
+        1.0,
+        1.0
+      ]
+    },
+    "terrain": {
+      "terrain_ambient": [
```
BenjamenMeyer, line 124: no need to duplicate the parent key in the child key.

engine.json
```diff
@@ -0,0 +1,1742 @@
...
+                "base_max_width__betternot": "3200",
+                "base_max_height__betternot": "2160",
+                "x_resolution": "3840",
+                "y_resolution": "2160",
+                "aspect": "1.7",
+                "font_point": "32"
+            }
+        },
+        "Text": {
+            "text_bitmap": {
```
BenjamenMeyer, line 225: just `bitmap`/`vector`/`antialiased_vector` is fine

engine.json
```diff
@@ -0,0 +1,1742 @@
...
+                "font_antialias": "false"
+            },
+            "text_aa_vector": {
+                "high_quality_font": "false",
+                "high_quality_font_computer": "false",
+                "font_antialias": "true"
+            }
+        },
+        "Geometry": {
+            "GeomRetro": {
```
BenjamenMeyer, line 242: `Retro`/`Low`/`Medium`/`High`/`VeryHigh`/`Extreme` would be be better

BenjamenMeyer (commented):
1. don't be afraid to add layers; they help bring clarity
2. don't be afraid to change names to things more meaningful - it'd be much appreciated
3. no need to repeat things - if text repeats between keys then it's probably something that should make another layer.

Yes, I know this will break things. That's okay.

engine.json
evertvorster: Not touching that, the engine expects this format.

engine.json
evertvorster: Not touching that, the engine expects this format.

theme.json
evertvorster: Not touching that, the engine expects this format.

theme.json
evertvorster: It could be, but that would require a change to the engine. I just want it to read the same variables from a different file so that we can start to retire the old file. We can always come back and mop up a little after the big demolition work is done.
For now, I don't want to change anything in this layout.

theme.json
evertvorster: Yes, it would be nicer. And we can get to that, but for now the engine expects this key. I don't want to change the keys the engine reads now, I want it to stop reading them from vegastrike.config. No little side quests into the code.

evertvorster:
> 1. don't be afraid to add layers; they help bring clarity
> 2. don't be afraid to change names to things more meaningful - it'd be much appreciated
> 3. no need to repeat things - if text repeats between keys then it's probably something that should make another layer.
> 
> Yes, I know this will break things. That's okay.

It being OK is debatable. Every change to a name, variable or layer here is a little side-quest that wastes time. It involves a tour into the engine internals, which then kicks up another bunch of things that might need fixing. 

No, I'm not falling into that trap.

The whole goal here is to get VS off having two separate and incompatible config systems. Its been in limbo for more than two years, and this is actively damaging the project. No one wants to play something that is visibly buggy and hard if not impossible to configure. 

I want to help the project get to a place where a new player can install VS, open a config app and fully config their game, and start playing. 

Once that is done, we can come back and move things around... which would be _easier_ as there would be only one config system to worry about. 

My overarching plan here is:
1. Get the engine to read split config files.  
2. Merge the split config files 
3. Swap over engine reading variables from vegastrike.config to config.json and friends
4. Remove vegastrike.config, and the code that read it
5. Upgrade the config utility
6. Put the config utility with the engine 

This is a fairly easy plan to follow, unless we start poking around in the 1400 settable things in VS. The config files that I want to merge has been populated with the exact same variables names and options that the engine reads today, so swapping them over is a matter of looking for the variable name and getting the setting in the new file.

theme.json
BenjamenMeyer: fair enough

BenjamenMeyer:
@evertvorster sounds good - changes look good; are we ready for testing/full review?

evertvorster:
> @evertvorster sounds good - changes look good; are we ready for testing/full review?

Yes. This is what I want to merge. It must be tested against the accompanying engine PR, as that is able to read and merge configs from different files. (and single files)

So, this is not getting merged before https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1719 is merged.

BenjamenMeyer:
@evertvorster can you confirm if this is from you or not? - notice the `inverse X` option is too high and conflicts with the menu bar across the top so you can barely read it.
image

BenjamenMeyer (approved):
there's an issue in the controls screen, but I don't think that's part of this PR - if it is, then please fix it; if not we can get that in a follow-up.
Not sure I'd remove the `Advanced` tab... but okay.

evertvorster:
> @evertvorster can you confirm if this is from you or not? -  The image is mine, but not related to tihs PR.  


I left everything the current config utility reads in the config.json. The stuff in the advanced tab we can bring in, but it should be on a one-by-one basis, and the current engine does not care where a setting is defined, as it merges the files before processing. 

What I intend to do for config utility is to port over my SDL3 imgui settings utiliity that I have been developing on a different branch into the engine repo, rather than the Assets repo. My reasoning here is two-fold. 
1. The engine should be able to configure itself, no matter which assets are being used. 
2. If we are to merge the config with the engine, it needs to be in the same repo. 

Once it works properly, we can retire the settings utility in Assets completely. In short, I'm not touching the current settings app at all.

evertvorster:
> there's an issue in the controls screen, but I don't think that's part of this PR - if it is, then please fix it; if not we can get that in a follow-up. Not sure I'd remove the `Advanced` tab... but okay.

On futher inspection, yes, I chopped Advanced tab. The text overflowing on the config app is because the screen is not tall enough. Once you expand it vertically, the text is scrolled down. this is pre-existing behavior. I'll fix it in a follow-up

PR #272, opened 2026-08-18
Fix install split config

evertvorster:

Please answer the following:

Code Changes:

Issues:
- - **Installs** the new split files with cmake
- - Fixes layout bug in kivy config app.

Purpose:
- - fix a layout bug in the config app. 
- - Install the split config files properly
- What release is this for?
- -0.11
--0.11


evertvorster:
Yeah, so that layout bug was caused by me. The default frame shows the bottom of the text, and so the top overlays whatever is above it. When you drag the window down, it becomes clear.  This PR changes the frame type, and now it does not overlay on the top, and you can scroll down. 

Also, I needed to actually **install** the split configs.

stephengtuggy (approved):
LGTM

PR #278, opened 2026-08-22
base_computer: read config via merged json (fix KeyError after config…

evertvorster:
… split)

The 2026-08-09 config split moved constants/components out of config.json into engine.json -> base. ship_view.py and upgrade_view.py still read config.json directly, so data['constants'] / data['components'] now raise KeyError and base screens break.

Add python/base_computer/config.py which merges the split config files the same way the engine (InitPaths) does: datadir default files then homedir overrides, config.json last in each group, engine.json's 'base' unwrapped so constants/components land at the merged root. The C++ engine's load_config does not unwrap 'base', which is why these keys were silently missing.

ship_view.py: query non_combat_mode_multiplier and megajoules_multiplier from the merged config; compute vsdm (kj_per_unit_damage/kilo) once and pass it to get_weapon_details, removing three hardcoded 5400 literals (a pre-existing display inconsistency vs upgrade_view's divide-by-kilo).

upgrade_view.py: query kj_per_unit_damage and kilo from the merged config.

Please answer the following:

Code Changes:

Issues:

Purpose:
- - fix more fallout from the split config files.
- What release is this for?
- -0.11
- -0.11

### DIFF
--- a/python/base_computer/config.py
+++ b/python/base_computer/config.py
@@ -1,0 +1,66 @@
+# Merge the split JSON config files the same way the engine does, so
+# base_computer scripts read the effective config instead of a single raw file.
+#
+# The engine (InitPaths) merges, in order, later values overwriting earlier:
+#   1. datadir bindings.json, theme.json, engine.json, config.json  (defaults)
+#   2. homedir bindings.json, theme.json, engine.json, config.json  (user overrides)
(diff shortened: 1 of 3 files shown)


stephengtuggy (approved):
Haven't tested it yet, but looks good to me! And good catch.

PR #280, opened 2026-08-23
Unify text coords

evertvorster:
**This PR depends on**: 
https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1755

Old PR that introduced this, which we are breaking up into digestible steps:
https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1737

It will break the current engine in various and interesting ways if run alone. 

Details of the implementation is in that PR, this is the accompanying changes that is needed in Assets for it all to work. 

Please answer the following:

Code Changes:

Issues:

Purpose:
- What release is this for?


bases/fixers/explore_beta.py
```diff
@@ -9,7 +9,7 @@
         quest.removeQuest(plyr,'gemini_sector/delta_navpoint',2)
     Base.Message('''Congratulations on your first mission! We have decided to go to the next system.
 Based on the data that you have collected, it appears that the other jump point should go
-to a system that we have named #55ffffbeta#000000. Again, collect info
+to a system that we have named #c.333:1:1#beta#-c. Again, collect info
```
royfalk: Hard disagree. Hex colors (e.g. FF0000) are an industry standard and as such, known by everyone. 

By moving to this, we are making it more difficult for artists to contribute. The engine should handle these elegantly.

controls.json
```diff
@@ -6,7 +6,7 @@
     "textColor": "0.1, 0.8, 0.1, 1.0",
     "rect": "-.96, .76, 1.9, .08",
     "color": "0.0, 0.0, 0.0, 0.0",
-    "font": "0.07, 1.75",
+    "font": "1.05, 1.75",
```
royfalk: Not sure what these numbers represent.

royfalk (commented):
Did not review the actual meat of this PR.
I assume we're moving from [-1.0 left, 0 center, 1.0 right] to [0.0 left, 0.5 center, 1.0 right].

royfalk:
Also https://github.com/vegastrike/Assets-Masters/issues/119, which you can easily do with your nifty AI and greatly improve the code.

evertvorster:
> Also [vegastrike/Assets-Masters#119](https://github.com/vegastrike/Assets-Masters/issues/119), which you can easily do with your nifty AI and greatly improve the code.

Will do, can it be folded in here, or would you like to see a separate PR?

bases/fixers/explore_beta.py
evertvorster: Accepted, reverting and changing course here to preserve the industry standard. The parser will still accept the new standard, as it allows some interesting things with alpha, but for now this will actually decrease the size of the PR quite a bit.

royfalk: > as it allows some interesting things with alpha

What do you mean? The hex format is already RRGGBB(AA), so alpha is supported but not required. 
Or is there something else you mean.

royfalk:
> > Also [vegastrike/Assets-Masters#119](https://github.com/vegastrike/Assets-Masters/issues/119), which you can easily do with your nifty AI and greatly improve the code.
> 
> Will do, can it be folded in here, or would you like to see a separate PR?

Separate PR. I think that one would generate a lot of interest. This sort of take a whole bunch of files and just move things around really a game changer in terms of velocity.

It used to be that it would take a lot of time to focus on polishing code, so we naturally stayed away from the python side. But a lot of code is there and needs attention. Finding duplicate code and refactoring. Documenting code so we'd know what can be done with what we have. 

Generating documentation for people of what's possible in English. You then say, I want you to create a new mission type X that does that and it just happens. **For me, this is perhaps the most important thing we've done since 2019.**

imgui.ini
```diff
@@ -0,0 +1,8 @@
+[Window][Debug##Default]
```
SGD1953: Was this intentionally committed?

bases/computer_lib.py
```diff
@@ -27,7 +27,9 @@ def time_key(filename):
 
 def savelist():
     global savefilters
-    return [ GUI.GUISimpleListPicker.listitem(path,path)
+    # Display the save name with underscores as spaces (e.g. "HeadingOut_2" ->
+    # "HeadingOut 2"), but keep the raw filename as the load data.
+    return [ GUI.GUISimpleListPicker.listitem(path.replace("_", " "), path)
```
SGD1953: Why are we not showing the savegame names as they are?

controls.json
SGD1953: Those are scale factors now, you probably want to look at https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1737 first

evertvorster: These are relative font-height scales on the engine's global font_point (1.0 = the default font height). The old 0.06–0.1 values were legacy absolute heights; we converted them to a multiplier so the base computer's font sizes scale consistently with everything else. 1.0 is normal, 0.9 smaller, 1.2 larger. (The second number is the stroke weight, unchanged.)

Its the same system the rest of the game now uses.

bases/fixers/explore_beta.py
evertvorster: I swapped back the colors to the industry standard as recommended, so these don't appear in this PR anymore.

bases/computer_lib.py
royfalk, line 32: I assume there's the opposite of this somewhere else?
Perhaps in https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1755:
> The save games screen had a bug where long saved game names with underscores in them were wrapping and overlapping. This has been fixed.

In any case, the old engine had and maybe still has code to deal with spaces and other weird filenames that were problematic on windows 98. 

This is of course no longer the case. I'm not sure we should do these conversions today. What happens if I have two files: 'save 1' and 'save_1'?
You could still randomly pick one but you won't know which is which.

Having said that, I've yet to review the engine PR that comes with this.

royfalk (approved):
OK. This looks pretty reasonable. 
One non-blocker noted.

bases/computer_lib.py
evertvorster: >What happens if I have two files: 'save 1' and 'save_1'?

Base computer does not support spaces in file names, you HAVE to type an underscore. When we change that, we can revisit this.

SGD1953: You can now modify the invalid chars in `controls.json`, just remove the space there and it will accept them.

royfalk:
Why was this closed? 
If you decide to close a PR, please add a reason why.

evertvorster:
Ah, yes. 

I realized we are still not in agreement with the coordinate system to be used, so I went back to the drawing board. 

I picked out the bugs that were standalone, and fixed them in separate PRs, and then also re-engineered the implementation process. 

So, thanks for the review, it really does help, and your comments shape the proper implementation, but this PR, as it stands can't be applied anymore.

PR #289, opened 2026-09-05
Upscaled and compressed cargo images with bc3 and improved consoles a…

SGD1953:
…nd holovids

Port from master, did spot check that the images are there


SGD1953:
@stephengtuggy here are the cargo images

stephengtuggy (approved):
LGTM!

